PR #22, opened 2016-08-31
Add :no_config message

slashmili:
add `:no_config` message for the cases that users forgot to configure `rustler_crates`


hansihe:
Thanks for the PR.

PR #29, opened 2016-11-07
add nif dirty func flags

### LINKED ISSUE #28: support for erlang's dirty scheduler?
Since rust in erlang is often used to do cpu intensive job, should rustler supports dirty scheduler? Otherwise rust calls will block erlang VM too long.

doomsplayer:
Fix #28


hansihe:
This looks good.

It would be nice if the `ErlNifTaskFlags` enum could be reexported by some other module, probably the root `rustler` module. Normally the user should neither import nor use the `wrapper` module directly.

doomsplayer:
@hansihe re-exported the flag in root module.

P.S. We are using this feature in production now.

hansihe:
Thanks for the contribution

PR #36, opened 2016-12-27
Use `.in.rs` extension for better syntax highlight support.

scrogson:


hansihe:
Change makes sense.

Seems like the build failed on something really stupid, namely installing the `erlang-mode` package? Because that package is only a `recommended` dependency of `erlang`, I think an easy fix would be to add `--no-install-recommends` to the `apt-get install` command in the travis dotfile.

PR #38, opened 2016-12-28
remove implicit inner RwLock in ResourceCell, require the Sync trait …

hansihe:
…instead


hansihe:
This breaks compatibility for the resource term API, any big problems?

scrogson:
👍

PR #39, opened 2016-12-28
Dynamic terms

hansihe:
Add support for term type checking.

### DIFF
--- a/src/atom.rs
+++ b/src/atom.rs
@@ -34,11 +34,11 @@ impl NifAtom {
     }
 }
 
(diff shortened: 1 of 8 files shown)


src/atom.rs
```diff
@@ -35,7 +35,7 @@ impl NifAtom {
 }
 
 pub fn is_term_atom(env: &NifEnv, term: NifTerm) -> bool {
```
scrogson: Should this be renamed to `is_atom` ?

hansihe: It should.

PR #42, opened 2016-12-30
`NifMapIterator`, `NifAtom::to_string`, and a few warning fixes

jorendorff:
The first 2 commits here just fix some warnings and fix a typo.

The rest are about `NifMapIterator` and `NifAtom::to_string`. I've been working with @scrogson on `fast_json` and we need these two features to be able to serialize Erlang maps as JSON.

### DIFF
--- a/src/atom.rs
+++ b/src/atom.rs
@@ -4,8 +4,17 @@ use std::sync::Mutex;
 use std::ops::DerefMut;
 
 use super::{ NifTerm, NifEnv };
-use ::wrapper::nif_interface::size_t;
-use ::wrapper::nif_interface::{ enif_make_atom_len, enif_alloc_env, NIF_ENV, NIF_TERM };
+use wrapper::nif_interface::{
+    ErlNifCharEncoding,
+    NIF_ENV,
+    NIF_TERM,
+    c_uint,
+    enif_alloc_env,
+    enif_get_atom,
+    enif_get_atom_length,
+    enif_make_atom_len,
+    size_t,
+};
 
 // Atoms are a special case of a term. They can be stored and used on all envs regardless of where
 // it lives and when it is created.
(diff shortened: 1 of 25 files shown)


src/atom.rs
```diff
@@ -32,6 +41,38 @@ impl NifAtom {
...
         }
     }
+
+    /// Get the contents of this atom as a string.
+    ///
+    /// (This is slow, and for most cases, unnecessary. For example, if you
+    /// need to know whether or not this atom is `true`, try comparing it to
+    /// `get_atom_init("true")`.)
+    ///
+    pub fn to_string(&self, env: &NifEnv) -> String {
```
hansihe: My implementation of this is a bit different. It is done completely without additional copying, but it allocates a full 257 bytes of memory regardless of how long the atom actually is. What do you think?

```rust
pub fn get_atom(env: NIF_ENV, term: NIF_TERM) -> Option<String> {
    // Max 257 characters can be written (256 + NULL).
    let mut string = String::with_capacity(257);

    {
        // We will validate UTF8 at the end of the function.
        let mut string_buf = unsafe { string.as_mut_vec() };

        let result = unsafe { nif_interface::enif_get_atom_latin1(env, term, string_buf.as_mut_ptr(), 257) };

        if result == 0 {
            return None;
        }

        // This SHOULD be safe unless the VM is lying to us.
        unsafe { string_buf.set_len((result - 1) as usize) };
    }

    // REQUIRED! This will verify that the data is actually valid
    // unicode without copying.
    match ::std::str::from_utf8(string.as_bytes()) {
        Ok(_) => Some(string),
        Err(_) => None,
    }
}
```

Note that my variant is untested, but it should mostly work.

scrogson:
👍

src/wrapper/map.rs
```diff
@@ -35,3 +35,41 @@ pub fn map_put(env: NIF_ENV, map: NIF_TERM, key: NIF_TERM, value: NIF_TERM) -> O
     }
     Some(result)
 }
+
+pub struct MapIterator {
```
hansihe: [The idea](https://github.com/hansihe/Rustler/blob/master/src/wrapper/mod.rs#L3-L6) behind the wrapper module was to have functions be as close as possible to the original C api, while being reasonably safe to use. I am therefore a bit reluctant to put a iterator implementation in here. What do you think?

src/atom.rs
hansihe: I should mention that I was working on this parallel with you.

src/wrapper/map.rs
jorendorff: Since this comment, I reimplemented it so that there's no Iterator implementation in wrapper/map.rs. I need to resolve conflicts, though.

src/atom.rs
jorendorff: Sorry I didn't see this last week.

Calling `enif_get_atom_length` is cheap; I'd do that rather than overallocate. You're right about the copy at the end of mine, though -- that's dumb. I'll make a separate PR and try to merge the two.

src/wrapper/map.rs
jorendorff: Oh, it turns out I still had this little newtype in there. But that will be easy to rip out.

hansihe:
Brilliant, thanks for the contribution!

PR #43, opened 2016-12-31
Term methods

hansihe:
Adds a lot of functions to NifTerm instead of having them be freestanding,

Do not merge yet.


hansihe:
One thing I have still not quite decided on is what the return types of methods on NifTerm should be.

Right now they return Option<_> because there is not much error information returned from the enif_* functions.

Would it make more sense to return Result<_, NifError> so that they can be used with try! directly?

PR #48, opened 2017-01-05
Rename scheduling flags

scrogson:
This PR gives more ergnomic use of NIF scheduling flags.

```rust
use rustler::schedule::NifScheduleFlags::*;

rustler_export_nifs! {
    "Elixir.MyNifModule",
    [("normal_nif", 1, normal_nif),
     ("dirty_cpu_nif", 1, dirty_cpu_nif, DirtyCpu),
     ("dirty_io_nif", 1, dirty_io_nif, DirtyIo)],
    Some(load)
}
```

### DIFF
--- a/src/schedule.rs
+++ b/src/schedule.rs
@@ -1,6 +1,12 @@
 use super::NifEnv;
 use super::wrapper::nif_interface::enif_consume_timeslice;
 
+pub enum NifScheduleFlags {
+    NORMAL = 0,
+    DIRTY_CPU = 1,
+    DIRTY_IO = 2,
+}
+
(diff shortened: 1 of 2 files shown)


src/schedule.rs
```diff
@@ -1,6 +1,12 @@
 use super::NifEnv;
 use super::wrapper::nif_interface::enif_consume_timeslice;
 
+pub enum NifScheduleFlags {
+    NORMAL = 0,
```
hansihe: I believe the general rust naming convention is to use CamelCase for enum variants. It's different in `erlang_nif_sys` because I belive it mirrors the naming in the C header.

scrogson: Ok

PR #50, opened 2017-01-06
`rustler::thread::spawn` for threaded NIFs.

jorendorff:
This also adds a public `env.error_tuple(reason)` method, just because it's handy.

### DIFF
--- a/src/thread.rs
+++ b/src/thread.rs
@@ -1,0 +1,58 @@
+use ::{ NifEnv, NifTerm, NifEncoder };
+use ::wrapper::nif_interface::{ self, ErlNifPid };
+use ::types::atom;
+use std::mem;
+use std::thread;
+use std::panic;
+
+/// Return the calling process's pid.
+fn caller(caller_env: &NifEnv) -> ErlNifPid {
+    let mut pid: ErlNifPid = unsafe { mem::uninitialized() };
+    unsafe {
+        nif_interface::enif_self(caller_env.as_c_arg(), &mut pid);
+    }
+    pid
+}
+
+/// Implements threaded NIFs.
+///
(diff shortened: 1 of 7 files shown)


src/thread.rs
```diff
@@ -0,0 +1,58 @@
+use ::{ NifEnv, NifTerm, NifEncoder };
+use ::wrapper::nif_interface::{ self, ErlNifPid };
+use ::types::atom;
+use std::mem;
+use std::thread;
+use std::panic;
+
+/// Return the calling process's pid.
+fn self_pid(env: &NifEnv) -> ErlNifPid {
```
scrogson: Should we just call this `caller`?

test/test/thread_test.exs
```diff
@@ -0,0 +1,51 @@
...
+      receive do
+        y -> y
+      after 1000 ->
+          :timeout
+      end
+    end)
+
+    # The OS scheduler guarantees virtually nothing about sleep().
+    # Answers may arrive out of order.
+    assert :lists.sort(results) == times
```
scrogson: can you `Enum.sort(results)`

src/thread.rs
jorendorff: No opinion, feel free to change it.

I made it non-public at the last minute because it didn't make sense having it inside a module named `thread`.

test/test/thread_test.exs
jorendorff: Yup!

src/thread.rs
scrogson:
```rust
fn caller(caller_env: &'a NifEnv) -> ErlNifPid { ... }
```

scrogson:
👍

scrogson:
❤️

PR #54, opened 2017-01-08
Make function calls to zero-arity functions explicit

NobbZ:
There was a warning introduced in Elixir 1.4 which complained about
zero-arity function calls without parentheses roughly like the
following:

> ```
> warning: variable "foo" does not exist and is beeing expanded to "foo()", please use parentheses to remove the ambiguity or change the variable name
> ```


hansihe:
Thanks!

PR #55, opened 2017-01-08
make thread::spawn more generic

hansihe:
This change would allow the user to use thread::spawn with things like thread pools. I think this may be a better solution than OwnedEnv because it enables us to easily implement term transfer to the other env.

cc: @scrogson @jorendorff

### DIFF
--- a/src/thread.rs
+++ b/src/thread.rs
@@ -15,6 +15,17 @@ fn caller<'a>(caller_env: NifEnv<'a>) -> ErlNifPid {
     pid
 }
 
+pub trait JobSpawner {
+    fn spawn<F: FnOnce() + Send + panic::UnwindSafe + 'static>(job: F);
+}
+
+pub struct ThreadSpawner;
+impl JobSpawner for ThreadSpawner {
+    fn spawn<F: FnOnce() + Send + panic::UnwindSafe + 'static>(job: F) {
+        thread::spawn(job);
+    }
+}
+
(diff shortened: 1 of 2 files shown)


test/src/test_thread.rs
```diff
@@ -11,7 +11,7 @@ pub fn threaded_fac<'a>(env: NifEnv<'a>, args: &Vec<NifTerm>) -> NifResult<NifTe
     }
 
     let n: u64 = args[0].decode()?;
-    thread::spawn(env, move |thread_env| {
+    thread::spawn::<thread::ThreadSpawner, _>(env, move |thread_env| {
```
hansihe: Is there a way to avoid the _ in <thread::ThreadSpawner, _>?

src/thread.rs
```diff
@@ -15,6 +15,17 @@ fn caller<'a>(caller_env: NifEnv<'a>) -> ErlNifPid {
     pid
 }
 
+pub trait JobSpawner {
```
scrogson: Maybe we call this `TaskSpawner`?

test/src/test_thread.rs
scrogson: maybe a type alias?

jorendorff: I don't see a way to avoid it while keeping `spawn` as its own function. It might make sense to offer it only as a method of the trait, though.

PR #60, opened 2017-01-10
`NifPid` and `env.send(pid, message)`

jorendorff:
The first commit adds our own `NifPid` wrapper type (like `NifTerm` and `NifEnv`) and stops exporting `ErlNifPid`.

The second commit adds an `env.send()` method.

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -3,6 +3,17 @@ use ::wrapper::nif_interface::{ self, NIF_ENV, NIF_TERM };
 use ::types::pid::NifPid;
 use std::sync::{Arc, Weak};
 
+impl<'a> NifEnv<'a> {
+    /// Send a message to a process.
+    pub fn send(self, pid: &NifPid, message: NifTerm<'a>) {
+        // Implementation note: As of ERTS 8.0 (Erlang/OTP 19), there is another way to do this:
+        // pass a null pointer to the `msg_env` parameter of `enif_send()`. That might be
+        // faster. For now, we copy the term into a process-independent environment, then send it.
+        OwnedEnv::new().send(pid, |env| message.in_env(env))
+    }
+}
+
+
 /// A process-independent environment.
 ///
 /// An owned environment is a place where Erlang terms can be created outside of a NIF call. Rust
--- a/src/env.rs
+++ b/src/env.rs
@@ -1,36 +1,18 @@
 use ::{ NifEnv, NifTerm };
-use wrapper::nif_interface::{ self, NIF_ENV, NIF_TERM };
-pub use wrapper::nif_interface::ErlNifPid;
-use std::mem;
+use ::wrapper::nif_interface::{ self, NIF_ENV, NIF_TERM };
+use ::types::pid::NifPid;
 use std::sync::{Arc, Weak};
-
-impl<'a> NifEnv<'a> {
-    /// Return the calling process's pid.
-    ///
-    /// # Panics
-    ///
-    /// Panics if this environment is process-independent.  (The only way to get such an
-    /// environment is to use `OwnedEnv`.  The `NifEnv` that Rustler passes to NIFs when they're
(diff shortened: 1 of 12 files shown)


jorendorff:
It stinks that we can't enforce the API rules without this thread-local variable. But I see no way to do it.

I measured on my laptop (Linux) and the overhead is measurable: 133ns, [about the same as a memory read that misses your CPU cache](https://people.eecs.berkeley.edu/~rcs/research/interactive_latency.html). Atomic operation. :-\

By comparison, calling the `echo_u8` NIF takes about 750ns to run, so the overhead is still small compared to the most trivial possible NIF.

We still have 2 options:
* take the thread-local variable; or
* ask the Erlang VM people to officially document that what we're doing is ok.

OvermindDL1:
> ask the Erlang VM people to officially document that what we're doing is ok.

It really should be documented at the very least yes or no, the mailing list would be the place to ask about the issue.

jorendorff:
(Stupid code I used to measure this, just in case I am doing something boneheaded. I just threw it into `primitives_test.exs`, because I am a troglodyte. But I'd love to know the right way to do this. :) )

```elixir
t0 = :os.system_time(:millisecond)
for _ <- 1..1000000 do
  RustlerTest.echo_u8(3)
end
t1 = :os.system_time(:millisecond)
IO.puts("elapsed time (msec):")
IO.puts(t1 - t0)
```

jorendorff:
@OvermindDL1 Which mailing list? erlang-questions?

OvermindDL1:
> @OvermindDL1 Which mailing list? erlang-questions?

Yep, the big authors check there where they do not check github.

hansihe:
One of the use-cases I first used  Rustler for was implementing a large mutable binary buffer. This involved many small NIF calls, which in many cases only did a memory read or write. I would like to keep rustler as usable as possible for those use-cases as well, and that is why I am hesitant about adding constant overhead.

That being said, we could certainly temporarily put this under a cargo feature flag. That way, people who need the feature can enable it and eat the overhead, while people who don't can leave it off. If this gets resolved with the erlang team, we can remove the feature flag. What do you think about this?

jorendorff:
The problem with making `OwnedEnv::send()` a configurable feature is that `rustler::thread::spawn` depends on it.

So I just made it unsafe.

(Experience makes me wary of configurable features anyway. Each one multiplies the number of times you need to compile just to make sure everything builds.)

src/env.rs
```diff
@@ -60,22 +60,18 @@ impl OwnedEnv {
...
-        where F: for<'a> FnOnce(NifEnv<'a>) -> NifTerm<'a>
-    {
-        let env_lifetime = ();
+    /// The environment is cleared as though by calling the `.clear()` method.
+    ///
+    /// # Unsafe
+    ///
+    /// This must not be called from a thread that is managed by the Erlang VM.
+    /// Also, `message` must be a valid term that belongs to this environment.
+    pub unsafe fn send(&mut self, recipient: &NifPid, message: NIF_TERM) {
```
hansihe: Would it make sense to use `enif_thread_type` here instead of having it be unsafe? I don't believe that function is very expensive.

scrogson: https://github.com/erlang/otp/blob/maint/erts/emulator/beam/erl_nif.c#L2719-L2735

scrogson: https://github.com/erlang/otp/blob/86d1fb0865193cce4e308baa6472885a81033f10/erts/emulator/beam/erl_nif.h#L212-L221

jorendorff:
OK, it uses `enif_thread_type` now and should be ready for another round of review!

I renamed `OwnedEnv::send` to `OwnedEnv::send_and_clear` since what it does is now quite different from `NifEnv::send`.

hansihe:
Finally merged!

PR #63, opened 2017-01-14
Keep moving toward "no crashes" promise.

jorendorff:
I decided to try to review all unsafe code in Rustler. I didn't get very far before seeing things I wanted to change.

1. `NifTerm::new()` is marked as safe, which is crazy. It's very unsafe. Unfortunately fixing this involves adding many `unsafe` blocks... but it is impossible to keep the "no crashes" promise if users can call this function in safe code.

2. I made a mistake in `SavedTerm::load(env)` that makes it possible to crash if you, like, move a `SavedTerm` to another thread and ... well, see the commit message for details. It's a race condition. Definitely a bug.

The other changes in this stack are pretty minor.


hansihe:
Good work!

PR #65, opened 2017-01-14
Compiler Improvements

scrogson:
This PR moves the compiler and new tasks into the appropriate directory: `lib/mix/tasks/` and names the files to match the conventions.

This PR also changes the compilation paths:

* Rust crates are now compiled into `_build/{Mix.env}/rustler_crates/{crate_name}/{rust_mode}/`
* The dynamic library is copied to `priv/native/` (Elixir takes care of symlinking this directory for us)


rustler_mix/lib/mix/tasks/compile.rustler.ex
```diff
@@ -35,10 +25,10 @@ defmodule Mix.Tasks.Compile.Rustler do
       |> make_no_default_features_flag(Keyword.get(config, :default_features, true))
       |> make_features_flag(Keyword.get(config, :features, []))
       |> make_build_mode_flag(build_mode)
-      |> make_platform_hacks(:os.type)
+      |> make_platform_hacks(:os.type())
 
     crate_full_path = File.cwd! <> crate_path
-    target_dir = Mix.Project.build_path <> "/rustler" <> crate_path
+    target_dir = Mix.Project.build_path() <> "/rustler_crates/#{id}"
```
scrogson: This could potentially just be called `crates` as well.

Issue #79, opened 2017-01-25
Improve usability


hansihe:
@jorendorff recently wrote a test project with rustler, and provided some usability feedback. We should see what we can do to remedy these issues.

1. It felt like too much manual configuration setup was needed.
2. Error handling was a pain.
3. Converting argument types, surprisingly, was a pain. `str` implements `NifEncoder`, but `String` and `&str` do not. This means you can't call `.encode()` on a tuple containing a string!
4. I tried making a resource for a type provided by the `image` crate and got a coherence error. Had to define a newtype. I don't know if there's anything we can do about that.
5. It seems like the type of `args` should be `&[NifTerm]`, not `&Vec<NifTerm>`.

Let me try to address this in order.

1. I definitely agree that getting up and running is way too hard at the moment. We should look at what we can do to improve the boilerplate generator, and we should try to write some guides.
2. I have felt bad about how error handling works for a while now. I am unsure what the best way to improve this would be, and feedback is definitely welcome in #2.
3. If there is no better way of doing it, we should write those additional implementations. Doing this feels like a workaround, but I am not sure if there is a better way of doing it.
4. I am aware of this, and have experienced it myself, but I am not sure if there is anything we can do to fix this.
5. Agreed. I seem to remember it was `&[NifTerm]` originally, and I am unsure when this was changed.

If there are anyone else having problems with usability, please post it in the comments. I would very much like to know where the pain-points are, so that we can look at addressing them.

jorendorff:
Item 4 is very hard to fix. The macros could define the trait, and then there would be no coherence worries. But that seems too silly to contemplate.

PR #80, opened 2017-01-25
Change type of NIFs' `args` parameter to `&[NifTerm<'a>]`.

jorendorff:
As mentioned in #79.


hansihe:
Could you change the type in the `rustler_mix` template as well?

PR #90, opened 2017-04-22
Fixed references to erlang_nif_sys in nif_interface.rs

kprieb:
Changed "self::erlang_nif_sys" to ":: erlang_nif_sys" in various spots in nif_interface.rs because the
latter, not the former, is the correct location of erlang_nif_sys module.  The former doesn't compile; the latter does.


hansihe:
Thank you for the pull request!

kprieb:
You're welcome; thank you for rustler!

jorendorff:
Any ideas why this compiled fine before on my machine and the test server?

Incidentally, paths in `use` are always absolute, so the leading `::` is unnecessary. We could write `pub use erlang_nif_sys::whatever;`.

kprieb:
Agreed, that is a puzzle.

PR #94, opened 2017-05-19
implement NifEncoder for Option<T> and Result<T,E>

alex-shapiro:
Implementation for built-in types:

`Option::Some(value)` is_encoded as `value_encoded`
`Option::None` is encoded as `:nil`
`Result::Ok(value)` is encoded as `{:ok, value_encoded}`
`Result::Err(err)` is serialized as `{:error, err_encoded}`

### DIFF
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -39,3 +39,21 @@ impl<'a, T> NifEncoder for &'a T where T: NifEncoder {
         <T as NifEncoder>::encode(self, env)
     }
 }
+
+impl<T> NifEncoder for Option<T> where T: NifEncoder {
+    fn encode<'c>(&self, env: NifEnv<'c>) -> NifTerm<'c> {
+        match *self {
+            Some(ref value) => value.encode(env),
+            None => atom::nil().encode(env),
+        }
+    }
+}
+
+impl<T, E> NifEncoder for Result<T, E> where T: NifEncoder, E: NifEncoder {
+    fn encode<'c>(&self, env: NifEnv<'c>) -> NifTerm<'c> {
+        match *self {
+            Ok(ref value) => (atom::ok().encode(env), value.encode(env)).encode(env),
+            Err(ref err) => (atom::error().encode(env), err.encode(env)).encode(env),
+        }
+    }
+}


jorendorff:
Want to add a test or two? It's easy - you can add a test NIF that encodes some Option/Result values in `test/src`; and use elixir to test the NIF (in `test/test`). To run the tests, `cd test; mix test`.

OvermindDL1:
The `Option::None` should probably be configurable.  Elixir may use the atom `:nil` as such a nil value in a lot of places (not everywhere), but the actual real `nil` value in the EVM/Erlang/BEAM is the empty list `[]`, it even has the name `nil` in the source and is used as the nil value through-out the erlang ecosystem.

hansihe:
@OvermindDL1 Since the interface of the NIF is probably consumed by the user, I think it would be best to just pick a value to standardize on for a null value. Since this is primarily aimed towards Elixir, I think the nil atom is a good choice.

Do you have a specific use-case for configuring the null value?

hansihe:
Going to merge this for now. We can add improvements/configurability later it it's needed.

OvermindDL1:
> Do you have a specific use-case for configuring the null value?

Interaction with Erlang libraries, which all use `[]` as their nil value.

OvermindDL1:
As well as it can be confusing because the Named type of NIL in the NIF api is for `[]`, getting an atom would be, highly unexpected and not match any docs on the subject at all.  I'd really recommend having `nil` matching the NIF nil (`[]`) and something like `nilatom` being the atom `nil`.

hansihe:
> Interaction with Erlang libraries, which all use [] as their nil value.

I understand what you're getting at, but I am having a hard time coming up with a case where a NIF function would interact directly with an erlang library.

As for `[]` being commonly used as a null value in the erlang ecosystem, I can't really find much documentation for that at all. The nif documentation refers to `[]` as `empty_list` only, never as nil or null. A google search for `erlang nil` recommends using the atoms `nil`, `null` or `undefined` as a null value, clicking at a couple of results, I couldn't find a single one recommending or documenting the use of `[]`. Looking through the erlang docs, I can find several functions using `undefined`, the function I could find using `[]` as a null value, mentioned in the documentation that this is strange behavior which is kept for historical reasons.

Sorry if it seems like I'm being petty here, but I would really like to avoid adding extra complexity if possible. 

How would you want it to be configurable? A cargo feature?

OvermindDL1:
@hansihe 

The nil type is named here, as well as it is called nil inside the engine code as well, among other areas: 
 http://erlang.org/doc/apps/erts/erl_ext_dist.html#id100749  :-)

> Sorry if it seems like I'm being petty here, but I would really like to avoid adding extra complexity if possible.

Oh nothing of the sort!  I'm just a bit of a stickler in regards to naming things, it can be paralyzing at times.  ^.^;

I do use `[]` as the 'nil' value in my elixir libraries though, it is just ever so slightly faster to compare than an atom is.  I honestly think elixir made a large mis-step by making `nil` === `:nil` instead of making `nil` === `[]` or so.  If people wanted `:nil` they could just write `:nil` (irritatingly stupid ruby heritage stuff...).

> How would you want it to be configurable? A cargo feature?

Honestly I'd just have them be two different types, I could easily see myself using both in a single project.

hansihe:
@OvermindDL1 Right, it is absolutely true that it's called nil internally and in the external term format. The thing is, it doesn't seem to me like it is actually used as a nil value anywhere.

I am not so sure an atom takes longer to compare than a `[]`. They are both represented as immediate Eterms, so they should both just require a simple integer equality check. Am I mistaken here?

Whether elixir made a mistake or not by making nil an atom is a discussion for another day, but as far as I can tell, the usage of `[]` as a nil value is not common in nether the erlang nor elixir ecosystems.

> Honestly I'd just have them be two different types, I could easily see myself using both in a single project.

That's not really possible in this instance. This is a pull request for a trait implementation of `NifEncoder` for `Option` and `Result`. There isn't really any way to make this configurable without having a cargo feature for it.

hansihe:
> For direct comparisons yes, atoms are interned into the type format directly, just as nil is, rather it is in a match context (when comparing the output to a set of values) where the speed benefit comes

Makes sense.

> There is no way to have a new return type instead of using the standard Option but rather a new enum based on Result or so where Error is always a [] and the Ok is whatever the user wants?

I don't quite understand what you mean. Are you suggesting we define an alternative to `std::option::Option` which encodes/decodes `None` to `[]`? By defining a custom enum instead of using the std one, you lose a lot of ergonomics. At that point I would think it would be easier to just do `option.ok_or(empty_list_term).encode(env)`.

OvermindDL1:
> I don't quite understand what you mean. Are you suggesting we define an alternative to std::option::Option which encodes/decodes None to []? By defining a custom enum instead of using the std one, you lose a lot of ergonomics. At that point I would think it would be easier to just do option.ok_or(empty_list_term).encode(env).

Basically, and I was worried about the ergonomic loss but was hoping a helper or macro would be able to simplify it.

OvermindDL1:
It's not a big deal at all, always a verbose work-around, it is just a case I'd use often so it would be nice to see simplified.  ^.^

OvermindDL1:
More thinking, I guess Elixir uses `:nil` when Erlang uses `undefined` or `[]`, it unifies the two cases I guess.  `undefined` is often used in Erlang to signify no return, where `[]` is often used in Erlang to signify 'work done but no response', hmm...

PR #96, opened 2017-05-22
implement NifDecoder for Option<T> and Result<T,E>

alex-shapiro:
The flip-side of #94, with tests for decoding and encoding.


hansihe:
Thanks!

Issue #97, opened 2017-05-22
Encoding and decoding Elixir structs


alex-shapiro:
The compiler plugin `ExStruct` for [encoding and decoding Elixir structs](https://docs.rs/rustler/0.14.0/rustler/types/elixir_struct/index.html) looks like it no longer exists in the codebase (except in an unused [test file](https://github.com/hansihe/rustler/blob/master/rustler_codegen/test.rs#L1997) in codegen). How easy would it be to modify `NifMap` to create a derive-able`NifStruct` macro? I'm imagining something like this:

```rust
#[derive(NifStruct(module = "MyStruct"))]
pub struct MyStruct {
    id: i32,
    name: String,
}
```

I've never written a derive-able macro before so I don't know how much effort this would take.

hansihe:
Fixed by #98

PR #98, opened 2017-05-23
Encode and decode Elixir structs

alex-shapiro:
Implements #97: a procedural macro that generates `NifEncode` and `NifDecode` implementations to convert between a Rust struct and an Elixir struct. For example, given the following definition:

```rust
#[derive(NifStruct)]
#[module = "MyStruct"]
pub struct MyStruct {
    id: i32,
    name: String,
}
```

`MyStruct{id: 123, name: "Bob Jefferson".to_owned()}` encodes as 
`%MyStruct{id: 123, name: "Bob Jefferson"}`.


hansihe:
Thanks for the PR!

PR #99, opened 2017-05-24
Attempt at fixing mix compiler on Windows

hansihe:
@tridius Does this fix your issue?


tridius:
@hansihe Just got a chance to try it out, and it looks like everything is building successfully. Thanks a lot for the quick response!

PR #108, opened 2017-07-31
Fix for working with umbrella projects

costaraphael:
I was searching the Elixir repo for some tips on making the compile task work with umbrella projects.

It turns out you just have to add the `@recursive true` attribute in the task module. :sweat_smile:

I tested a little bit and it seems that my PR does not break individual projects compilation.


hansihe:
Thanks!

PR #114, opened 2017-12-15
Encode and decode records (tagged tuples)

evnu:
This changeset adds functionality to derive a Rust struct from an Erlang [record](http://erlang.org/doc/reference_manual/records.html), which could also be generated by a [macro](https://hexdocs.pm/elixir/Record.html) in Elixir. Its advantage over decoding an Elixir struct is reduced runtime overhead, as no maps are involved and tuple elements can be accessed in constant time. Its advantage over decoding directly from a tuple is that it performs a simple check to verify that the record's tag matches the one described with the `tag` attribute of the rust struct. The code merges pre-existing functionality from `NifStruct` and `NifTuple`.

Consider this example taken from the test case. We can define a record using Elixir:

```elixir
defmodule AddRecord do
  import Record
  defrecord :record, [lhs: 1, rhs: 2]
end
```

With this changeset, this record can be represented with rustler:

```rust
#[derive(NifRecord)]
#[tag = "record"]
struct AddRecord {
    lhs: i32,
    rhs: i32,
}

pub fn record_echo<'a>(env: NifEnv<'a>, args: &[NifTerm<'a>]) -> NifResult<NifTerm<'a>> {
    let record: AddRecord = args[0].decode()?;
    Ok(record.encode(env))
}
```

### DIFF
--- a/rustler_codegen/src/lib.rs
+++ b/rustler_codegen/src/lib.rs
@@ -11,6 +11,7 @@ extern crate quote;
 
 mod util;
 mod tuple;
+mod record;
 mod map;
 mod ex_struct;
 mod unit_enum;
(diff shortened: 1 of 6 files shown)


evnu:
Note that the Travis CI tests failed with `** (Mix) The task "local.hex" could not be found`.

hansihe:
Sorry for the late reply, this seems good to me.

Thanks for the PR!

evnu:
Thank you!

PR #116, opened 2018-01-10
Remove Nif prefix from types

scrogson:
This PR is a work in progress to remove the `Nif` prefix from the types.

This should be in working order already but there are still more types to process.

### TODO

- [x] `NifEncoder` => `Encoder`
- [x] `NifDecoder` => `Decoder`
- [x] `NifTerm` => `Term`
- [x] `NifEnv` => `Env`
- [x] `NifError` => `Error`
- [x] `NifAtom` => `Atom`
- [x] `NifBinary` => `Binary`
- [x] `NifPid` => `Pid`
- [x] `NifListIterator` => `ListIterator`
- [x] `NifMapIterator` => `MapIterator`
- [x] `NifResourceType` => `ResourceType`
- [x] `NifResourceTypeProvider` => `ResourceTypeProvider`
- [x] `NifScheduleFlags` => `SchedulerFlags`
- [ ] ~~Type aliases for backwards comptatibility~~


hansihe:
I fully agree with the change, this is something that should have been done a while ago.

A couple of notes:
* Do you think it would be a good idea to add type aliases to minimize breakage?
* The example the Readme would need to be updated
* In a PR I just merged, there is another derivable encoder/decoder. This should probably be renamed as well.

scrogson:
> Do you think it would be a good idea to add type aliases to minimize breakage?

Yes, absolutely!

> The example the Readme would need to be updated

Taken care of.

> In a PR I just merged, there is another derivable encoder/decoder. This should probably be renamed as well.

Also taken care of.

scrogson:
Unfortunately, trait aliases haven't been fully implemented yet:

```
error: trait aliases are not yet fully implemented (see issue #41517)
  --> src/lib.rs:75:1
   |
75 | pub trait NifDecoder<'a> = Decoder<'a>;
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: add #![feature(trait_alias)] to the crate attributes to enable
```

I attempted to be a good citizen and add type aliases for everything but unfortunately, I think we're going to need to bite the bullet and break the API. We can provide an upgrade guide, which should be pretty minimal.

scrogson:
@hansihe regarding the custom derives...should I revert the changes to those since they aren't namespaced?

Should we name those `Rustler<Type>` or `Erl<Type>` rather than `Nif<Type>`?

Examples

```rust
#[derive(RustlerMap)]
```
or

```rust
#[derive(ErlMap)]
```

or should we just leave it the same?

hansihe:
@scrogson I personally think we should just leave it the same. I am mostly neutral between the three different naming schemes, so I land on keeping it the same.

As for how this should be released, I think there should be at least one release supporting the latest OTP version with the old API, then we can immediately release a new version with the new API afterwards.

mischov:
Big agreement about the "latest OTP version with old API then new API" suggestion.

scrogson:
@hansihe ok cool. I'll work on those changes.

> As for how this should be released, I think there should be at least one release supporting the latest OTP version with the old API, then we can immediately release a new version with the new API afterwards.

:+1:

PR #129, opened 2018-02-07
Delete preexisting dynamic library first

### LINKED ISSUE #128: Segfault in other process on recompilation
In `Mix.Tasks.Compile.Rustler.compile_crate/1`, Rustler uses `File.cp!/2` to copy a generated library to its destination. On my Linux, this will produce a segfault in another beam process if that process is using the library already.

## Example

Consider Rustlers default example after initializing Rustler.

```
$ mix rustler.new
This is the name of the Elixir module the NIF module will be registered to.
Module name > Example
This is the name used for the generated Rust crate. The default is most likely fine.
Library name (example) > example
```

Wire the generated files up as indicated in `native/example/README.md`. Then, add a runner:

```elixir
defmodule SegfaultingRustler do
  @moduledoc """
  Documentation for SegfaultingRustler.
  """

  def run do
    IO.puts "...Running..."
    run1()
  end

  defp run1 do
    {:ok, 3} = Example.add(1, 2)
    run1()
  end
end
```

Open a shell and run this example:

```
$ mix run -e SegfaultingRustler.run
...Running...
```

Now, open a second shell and recompile:

```
$ mix compile
```

On my system, this reliably results in a segfault.

## Cause

Dynamic libraries are mmapped into the target process and bound to the libraries inode. By using `cp`, something bad happens with that memory region.  I am not sure if it is invalidated, or somehow overwritten. In any case, this will produce a segfault.

## Solution

It is important to not meddle with the mmapped library's inode. Deleting the file before copy works here. If I run the example above, then `rm` the library and recompile, nothing bad happens.

### My Environment

Linux mybox 4.14.15-1-ARCH #1 SMP PREEMPT Tue Jan 23 21:49:25 UTC 2018 x86_64 GNU/Linux

`Cargo.toml`:
```
...
"checksum rustler 0.15.1 (registry+https://github.com/rust-lang/crates.io-index)" = "c9c618866bb92c22a30810b593809a41ee65640d82bca664954adcc2ed0a1cef"
"checksum rustler_codegen 0.15.1 (registry+https://github.com/rust-lang/crates.io-index)" = "d7627a6cea2d3aa1146eeed1f99c794f2dc01ab7b7072896a1561ed730bb024d"
...
```

### GDB

I compiled OTP with debug symbols to produce a  backtrace of the example above. With kerl:

```
CFLAGS="-g -O0" kerl build 20.2 20.2-debug-O0
```

```
$ coredumpctl gdb <core dump id>
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
Core was generated by `/home/evnu/kerl/20.2-debug-O0/erts-9.2/bin/beam.smp -- -root /home/evnu/kerl/20.2-d'.
Program terminated with signal SIGBUS, Bus error.
#0  0x00007f8edaadcc95 in rustler::types::primitive::<impl rustler::types::NifDecoder<'a> for i64>::decode () from /home/evnu/work/evnu/experiments/segfaulting_rustler/_build/dev/lib/segfaulting_rustler/priv/native/libexample.so
[Current thread is 1 (Thread 0x7f8f0d37f700 (LWP 14962))]
(gdb) bt
#0  0x00007f8edaadcc95 in rustler::types::primitive::<impl rustler::types::NifDecoder<'a> for i64>::decode () from /home/evnu/work/evnu/experiments/segfaulting_rustler/_build/dev/lib/segfaulting_rustler/priv/native/libexample.so
#1  0x00007f8edaad89f7 in example::add () from /home/evnu/work/evnu/experiments/segfaulting_rustler/_build/dev/lib/segfaulting_rustler/priv/native/libexample.so
#2  0x00007f8edaad94a0 in std::panicking::try::do_call () from /home/evnu/work/evnu/experiments/segfaulting_rustler/_build/dev/lib/segfaulting_rustler/priv/native/libexample.so
#3  0x00007f8edab1e63f in panic_unwind::__rust_maybe_catch_panic () at /checkout/src/libpanic_unwind/lib.rs:101
#4  0x00007f8edaada068 in rustler::codegen_runtime::handle_nif_call () from /home/evnu/work/evnu/experiments/segfaulting_rustler/_build/dev/lib/segfaulting_rustler/priv/native/libexample.so
#5  0x000000000044dea6 in process_main (x_reg_array=0x7f8f0f7400c0, f_reg_array=0x7f8f0f742140) at beam/beam_emu.c:3601
#6  0x0000000000537118 in sched_thread_func (vesdp=0x7f8f0e342e80) at beam/erl_process.c:8892
#7  0x00000000006f6157 in thr_wrapper (vtwd=0x7ffe31a1c0c0) at pthread/ethread.c:118
#8  0x00007f8f5042e08c in start_thread () from /usr/lib/libpthread.so.0
#9  0x00007f8f4ff5de7f in clone () from /usr/lib/libc.so.6
```

Note that I have seen crashes in `example::nif_init::FUN_ENTRIES::nif_func ()` as well, which remind of the stacktrace present in #113.

evnu:
Fix #128. Copying the build library on top of a preexisting build can result in a segfault if a node is currently using the already-built library.

### DIFF
--- a/rustler_mix/lib/mix/tasks/compile.rustler.ex
+++ b/rustler_mix/lib/mix/tasks/compile.rustler.ex
@@ -64,6 +64,11 @@ defmodule Mix.Tasks.Compile.Rustler do
     compiled_lib = Path.join([target_dir, Atom.to_string(build_mode), src_file])
     destination_lib = Path.join(priv_dir(), dst_file)
 
+    # If the file exists already, we delete it first. This is to ensure that another
+    # process, which might have the library dynamically linked in, does not generate
+    # a segfault. By deleting it first, we ensure that the copy operation below does
+    # not write into the existing file.
+    File.rm(destination_lib)
     File.cp!(compiled_lib, destination_lib)
   end
 


pichi:
There is common trick how to perform an atomic operation on *nix filesystem. If you copy into a temporary file at the same volume first and then move temporary to a destination. Then move operation is an atomic operation which means kernel performs unlink and move at once. There is not a time window where the file is missing or incomplete. I would suggest this approach to use here.

evnu:
@pichi do you know if `:file.rename/2` from erlang maps to atomic mv? I did not check that function, as I assumed that atomicity is not an issue here.

pichi:
I'm not familiar with Elixir enough to know how to make random file name so you should fill it.
```elixir
compiled_lib = Path.join([target_dir, Atom.to_string(build_mode), src_file]) 
destination_lib = Path.join(priv_dir(), dst_file)
tmp_lib = Path.join(priv_dir(), random_name())

# Perform atomic copy
File.cp!(compiled_lib, tmp_lib)
File.rename(tmp_lib, destination_lib)
```

I hope `File.rename` is implemented using glibc `rename()` or `mv` command which does it correctly. I don't see `File.rename!` variant which makes me a little bit nervous.

pichi:
@evnu I'm not sure about `:file.rename/2`. I will see.

pichi:
Yes, `:file.rename/2` for `prim_file` can be tracked down to
https://github.com/erlang/otp/blob/76c4821cff1e91c26dd2f812fc8cb40aa7d48e57/erts/emulator/nifs/unix/unix_prim_file.c#L813
seems OK.

evnu:
Elixir's `File.rename` delegates to `file:rename/2`:

```elixir
# lib/elixir/lib/file.ex
  def rename(source, destination) do
    :file.rename(source, destination)
  end
```

I am unsure how to generate a random file name without falling back to `mktemp`. I did not find functions which are guaranteed to generate unique values across different nodes. `make_ref/0` does not produce unique references across instances, and getting the time also seems to allow race coditions. Ideas?

Note that the proposed change in https://github.com/hansihe/rustler/pull/129#issuecomment-363755568 might leave temporary files around on some systems. If moving a file onto an opened file is not allowed on a certain operating system (I bet this is the case on Windows), the temporary file might not be removed. The original implementation in this PR fails gracefully if deleting an opened file is not allowed, as it does not generate additional files in `priv/`.

PR #134, opened 2018-02-23
Fix return value for compiler `run` callback on `prod` environment

### LINKED ISSUE #133: Warning generated during compilation
When I execute ` MIX_ENV=prod mix compile` I get the following warning generated:
`[warning] Mix compiler :rustler was supposed to return :ok | :noop | {:ok | :noop | :error, [diagnostic]} but it returned [<file list>]`

The docs about a compiler's `run/1` callback say that:

_Receives command-line arguments and performs compilation. If it produces errors, warnings, or any other diagnostic information, it should return a tuple with the status and a list of diagnosticsReceives command-line arguments and performs compilation. If it produces errors, warnings, or any other diagnostic information, it should return a tuple with the status and a list of diagnostics_

But the last line of the rustler compiler is:
``` Elixir
    # Workaround for a mix problem. We should REALLY get this fixed properly.
    _ = symlink_or_copy(config,
      Path.expand("priv"),
      Path.join(Mix.Project.app_path(config), "priv"))
```

If I add an `:ok` at the end, the warning goes away. Is this the correct fix?

cristianberneanu:
Closes #133

### DIFF
--- a/rustler_mix/lib/mix/tasks/compile.rustler.ex
+++ b/rustler_mix/lib/mix/tasks/compile.rustler.ex
@@ -144,6 +144,7 @@ defmodule Mix.Tasks.Compile.Rustler do
         File.rm_rf!(target)
         File.cp_r!(source, target)
       end
+      :ok
     else
       Mix.Utils.symlink_or_copy(source, target)
     end


scrogson:
❤️

PR #136, opened 2018-03-24
Move Env and Error to modules

scrogson:
@hansihe what do you think about this? I think it cleans things up a bit.


hansihe:
Yep, agreed.

PR #142, opened 2018-05-09
Add compilation support for FreeBSD

walkr:
Update make_lib_name to support FreeBSD

### DIFF
--- a/rustler_mix/lib/mix/tasks/compile.rustler.ex
+++ b/rustler_mix/lib/mix/tasks/compile.rustler.ex
@@ -106,6 +106,7 @@ defmodule Mix.Tasks.Compile.Rustler do
       {:win32, _} -> {"#{base_name}.dll", "lib#{base_name}.dll"}
       {:unix, :darwin} -> {"lib#{base_name}.dylib", "lib#{base_name}.so"}
       {:unix, :linux} -> {"lib#{base_name}.so", "lib#{base_name}.so"}
+      {:unix, :freebsd} -> {"lib#{base_name}.so", "lib#{base_name}.so"}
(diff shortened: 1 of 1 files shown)


rustler_mix/lib/mix/tasks/compile.rustler.ex
```diff
@@ -106,6 +106,7 @@ defmodule Mix.Tasks.Compile.Rustler do
       {:win32, _} -> {"#{base_name}.dll", "lib#{base_name}.dll"}
       {:unix, :darwin} -> {"lib#{base_name}.dylib", "lib#{base_name}.so"}
       {:unix, :linux} -> {"lib#{base_name}.so", "lib#{base_name}.so"}
+      {:unix, :freebsd} -> {"lib#{base_name}.so", "lib#{base_name}.so"}
       # {:unix, _} -> Assume .so? Is this a unix thing?
```
scrogson: I feel like maybe we should just make this the catch all for `unix` based systems other than `darwin`.

scrogson: @walkr I am certainly not 100% sure that is how *all* unixes work, but it's probably better at this point to just assume that `darwin` is the only exception. If we find that it's different in the future, we can always add a special case for it.

Can you just update this to be:

```ex
{:win32, _} -> {"#{base_name}.dll", "lib#{base_name}.dll"}
{:unix, :darwin} -> {"lib#{base_name}.dylib", "lib#{base_name}.so"}
{:unix, _} -> {"lib#{base_name}.so", "lib#{base_name}.so"}
```

Thanks!

walkr: @scrogson Done!

PR #146, opened 2018-06-04
Use workspaces

scrogson:
This PR updates the repo to use cargo [workspaces](https://doc.rust-lang.org/book/second-edition/ch14-03-cargo-workspaces.html)


scrogson:
@hansihe great question!

So it appears to work seamlessly (very similar to how it works today).

I tested this out here: https://github.com/scrogson/beamcoin/compare/workspaces and it works fine.

PR #155, opened 2018-06-16
Added documentation on codegen functions.

kraythen:
Added a bunch of user documentation to the codegen crate. Will prevent a user from having to troll through the source and test code to figure stuff out.


hansihe:
Thank you!

Issue #145, opened 2018-06-04
undefined reference enif_snprintf and others


evnu:
The following example results in linker errors after compilation. Rustler 0.16.0, rustc 1.26.0. It is unclear if this is caused within Rustler, or if it somehow stems from `erlang_sys-nif`. Any ideas are appreciated. I specifically encountered a missing `enif_snprintf` reference, which I tracked down to be caused from `NifAtom`.

```rust
#[macro_use]
extern crate rustler;

use rustler::env::OwnedEnv;
use rustler::types::atom::NifAtom;

fn main() {
    OwnedEnv::new().run(|env| {
        if let Ok(a) = NifAtom::from_str(env, "hello") {
            if let Ok(b) = NifAtom::from_str(env, "hello") {
                assert_eq!(a, b)
            }
        }
    });
}
```

Cargo.toml:

```toml
[package]
name = "rustler_enif_snprintf_unresolved"
version = "0.1.0"
authors = []

[dependencies]
rustler = "0.16"
rustler_codegen = "0.16"
lazy_static = "0.1"
```

Error Message:

```
/tmp/rustler_enif_snprintf_unresolved/target/debug/deps/librustler-0dc2f9159eb90b83.rlib(rustler-0dc2f9159eb90b83.rustler9.rcgu.o): In function `<rustler::env::OwnedEnv as core::ops::drop::Drop>::drop':
          /home/mo/.cargo/registry/src/github.com-1ecc6299db9ec823/rustler-0.16.0/src/env.rs:186: undefined reference to `enif_free_env'
          /tmp/rustler_enif_snprintf_unresolved/target/debug/deps/librustler-0dc2f9159eb90b83.rlib(rustler-0dc2f9159eb90b83.rustler11.rcgu.o): In function `rustler::wrapper::nif_interface::enif_alloc_env':
          /home/mo/.cargo/registry/src/github.com-1ecc6299db9ec823/rustler-0.16.0/src/wrapper/nif_interface.rs:59: undefined reference to `enif_alloc_env'
          /tmp/rustler_enif_snprintf_unresolved/target/debug/deps/librustler-0dc2f9159eb90b83.rlib(rustler-0dc2f9159eb90b83.rustler11.rcgu.o): In function `rustler::wrapper::nif_interface::enif_make_atom_len':
[...]

PR #156, opened 2018-06-17
WIP: Integrate cargo test

### LINKED ISSUE #147: Integrate cargo tests into Elixir test build.
It would be great if we could integrate cargo tests into the mix test task. If this is already there, pardon me but I have looked for docs.

evnu:
This implements `mix rustler.test` to fix #147. It does so by calling into the cargo API from within a NIF. The implementation is a little rough around the edges: 

* The crate is compiled within the NIF (by cargo itself) and ends up in `native/<crate>/target`. This should be fixable by setting the build path differently (I am using defaults right now) and maybe use `Mix.run("compile.rustler")` before executing the tests. Ideas and hints are welcome.
* Cargo is currently pulled in for all modes. It should only be needed when testing.

To test the implementation, run `mix rustler.test` within `rustler_mix/`. After some lengthy compilation, it should print this:

```
$ cd rustler_mix; mix rustler.test
Compiling NIF crate :rustler_cargo_test_runner (native/rustler_cargo_test_runner)...
    Finished release [optimized] target(s) in 0.0 secs
Compiling 1 file (.ex)
Running tests for rustler_cargo_test_runner

...snip compilation output...

running 1 test
test tests::success ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,8 @@ Cargo.lock
 /rustler_codegen/test/target/
 /rustler_codegen/Cargo.lock
 
+/rustler_mix/native/*/target
+
 /_build
 /cover
 /deps
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,5 +2,6 @@
 members = [
   "rustler",
   "rustler_codegen",
-  "rustler_tests"
+  "rustler_tests",
+  "rustler_mix/native/rustler_cargo_test_runner"
 ]
--- a/rustler_mix/lib/mix/tasks/rustler.ex
+++ b/rustler_mix/lib/mix/tasks/rustler.ex
@@ -1,0 +1,34 @@
+defmodule Mix.Tasks.Rustler do
+  use Mix.Task
+
+  @shortdoc "Prints Rustler help information"
+
+  @moduledoc """
+  Prints Rustler tasks and their information
+
+      mix rustler
+  """
+
+  @doc false
+  def run(args) do
+    {_opts, args, _} = OptionParser.parse(args)
+
+    case args do
+      [] -> general()
+      _ ->
+        Mix.raise "Invalid arguments, expected: mix rustler"
+    end
+  end
+
+  defp general() do
+    Application.ensure_all_started(:rustler)
+    Mix.shell.info(
+      """
+      Rustler v#{Application.spec(:rustler, :vsn)}
+      Safe Rust bridge for creating Erlang NIF functions.\n
+      Available tasks:
+      """
+    )
+    Mix.Tasks.Help.run(["--search", "rustler."])
+  end
+end
--- a/rustler_mix/lib/mix/tasks/rustler.test.ex
+++ b/rustler_mix/lib/mix/tasks/rustler.test.ex
@@ -1,0 +1,64 @@
+defmodule Mix.Tasks.Rustler.Test do
+  use Mix.Task
+
+  @shortdoc "Run cargo tests from Elixir"
+
+  @moduledoc """
+  This task runs `cargo test` for NIF crates configured to be used with Elixir.
+
+  By running Rust tests directly from within BEAM, users can also write tests in Rust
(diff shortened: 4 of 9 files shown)


evnu:
Regarding the failing travis build: The PR currently uses `compile.rustler` within `rustler_mix` itself. This only works if that task was already built before. To bootstrap this properly, this specific task might better be implemented in a separate elixir package (e.g. `rustler_cargo_test_runner`).

hansihe:
Thanks for the PR!

I'm a bit hesitant to merge a solution like this, I would prefer not to have a NIF in rustler_mix unless it's absolutely necessary.

Is there no way this could be made to use the cargo command line?

evnu:
You mean shelling out directly to cargo? I believe this can't work as the Erlang runtime needs to be present to link the NIF properly. But maybe somebody else has a better idea? In any case, I implemented this as part of `rustler_mix` for simplicity. A separate mix project (could be another git repo as well) would be possible to split this from `rustler_mix`.

scrogson:
In my mind, `mix rustler.test` would call `cargo test` for each of the configured crates in the `rustler_crates` in `mix.exs`.

The compile task in `rustler_mix` already calls out to `cargo` to compile crates. I not sure I understand what's different about running `cargo test` from a mix task?

evnu:
## Motivation

I did not explain the reasoning behind the NIF (and this PR in itself) properly: The motivation of 
providing a test runner as a NIF is to _circumvent linker errors_.

## Explanation

When running `cargo test`, functions such as `enif_snprintf()` cannot be used as part of the tests, as the linker cannot know where to find that function (see [here](https://github.com/hansihe/rustler/issues/145) for an example). As those `enif_X` symbols might be wrapped within Rustler, users will only see a confusing linker error. This is why I argue that shelling out to `cargo test` is not possible.

Now, when running the same tests from within BEAM itself, the symbols will be resolved when loading the NIF. Users can then use their `NifStruct` annotated structs in unit tests for example, which is not possible otherwise (again, due to `enif_snprintf()` not being resolved if the tests are not linked against BEAM).

> Is there no way this could be made to use the cargo command line?

I believe there is no other way. Maybe it would be possible to somehow resolve `enif_X` symbols differently, e.g. by finding a library which provides those functions or by mocking them. I think though that it is better to link "against the real thing" if possible.

> The compile task in rustler_mix already calls out to cargo to compile crates. I not sure I understand what's different about running cargo test from a mix task?

I hope the explanation above answers this question: `enif_X` can be resolved.

## Disclaimer

Please note that I am OK if you do not want to merge this attempt, I played around with it and found it interesting enough to propose it. A solution with less boilerplate and which does not depend on cargo would be preferable, but I did not find a workable way to solve the linker issues without running the tests from within BEAM.

## Example

To illustrate this further, here is the example from #145 as a unit test. If pasted that into the test module of the NIF from this PR to try it.

```rust
#[cfg(test)]
mod tests {
    use rustler::env::OwnedEnv;
    use rustler::types::atom::Atom;

    #[test]
    fn hello_world() {
        OwnedEnv::new().run(|env| {
            if let Ok(a) = Atom::from_str(env, "hello") {
                if let Ok(b) = Atom::from_str(env, "hello") {
                    assert_eq!(a, b)
                }
            }
        });
    }
}
```

When running this directly from `cargo test`, this fails to link (`undefined reference to enif_alloc_env'` and others). The example runs fine here with `mix rustler.test` from this PR.

evnu:
Is there any interest in further work for this PR? I would be up to put some more work in it, if it is useful for the project. Otherwise, I am ok with closing it.

evnu:
It has been a very long time without any activity here. I will close the PR.

dvic:
@evnu would you reconsider this PR? We stumbled upon this problem ourselves :) If not, I can open a PR where we document that running cargo test from a NIF project is not supported.

PR #159, opened 2018-06-21
support OTP-21

scrogson:


scrogson:
Looks like there's a segfault with Elixir 1.6 and OTP 21.0...

PR #165, opened 2018-07-15
add decoder for RangeInclusive

zacky1972:
sample code: 

```elixir
defmodule RustlerTest.RangeTest do
  use ExUnit.Case, async: true

  test "range iteration" do
    assert 55 == RustlerTest.sum_range(1..10)
  end
end
```

```rust
use rustler::{Env, Term, Encoder, NifResult};
use std::ops::RangeInclusive;

pub fn sum_range<'a>(env: Env<'a>, args: &[Term<'a>]) -> NifResult<Term<'a>> {
        let range: RangeInclusive<i64> = args[0].decode()?;

        let total: i64 = range.into_iter().sum();
    Ok(total.encode(env))
}
```

### DIFF
--- a/rustler/src/types/map.rs
+++ b/rustler/src/types/map.rs
@@ -2,6 +2,7 @@
 
 use wrapper::map;
 use {Decoder, Env, Error, NifResult, Term};
+use std::ops::RangeInclusive;
 
 pub fn map_new<'a>(env: Env<'a>) -> Term<'a> {
     unsafe { Term::new(env, map::map_new(env.as_c_arg())) }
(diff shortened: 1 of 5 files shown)


rustler/src/types/map.rs
```diff
@@ -182,3 +183,20 @@ impl<'a> Decoder<'a> for MapIterator<'a> {
...
     }
 }
+
+impl<'a, T> Decoder<'a> for RangeInclusive<T>
+where
+    T: Decoder<'a>,
+{
+    fn decode(term: Term<'a>) -> NifResult<Self> {
+        let vec:Vec<(Term, Term)> = term.decode::<MapIterator>()?.collect();
+        match (&*vec[0].0.atom_to_string()?, &*vec[0].1.atom_to_string()?) {
```
hansihe: This strikes me as fairly fragile, since erlang maps are unordered, our __struct__ key doesn't necessarily end up here. Might be a better idea to use the `Term.map_get` function here?

scrogson:
Thanks for the PR @zacky1972!

I've fixed up the merge conflicts and updated the code in #205. I've also preserved your commit so you get the attribution :+1:

PR #166, opened 2018-09-17
Add support for compiling Rust binaries

### LINKED ISSUE #78: Feature request: support for ports
There's [some code already for decoding ETF terms within Rust programs](https://github.com/seriyps/rust-erl-ext), which I'm hoping to use with Rustler (mostly relegating Rustler to compiling crates for Rust-implemented Erlang ports) in order to implement a port or two for an Elixir app (using [this guide](http://www.theerlangelist.com/article/outside_elixir) as a reference, albeit with Rust instead of Ruby).  Would be nice to have a bit tighter integration on this front, though (in particular, being able to run something like `mix rustler.port.new` or `mix rustler.new --port` or something to generate the boilerplate like with NIF crates).

If I or someone else were to implement something like this for Rustler, would a PR be welcome?  Or is this out of scope for Rustler?

cristianberneanu:
Closes #78

### DIFF
--- a/rustler_mix/lib/mix/tasks/compile.rustler.ex
+++ b/rustler_mix/lib/mix/tasks/compile.rustler.ex
@@ -28,13 +28,6 @@ defmodule Mix.Tasks.Compile.Rustler do
 
     Mix.shell.info "Compiling NIF crate #{inspect id} (#{crate_path})..."
 
-    compile_command =
-      make_base_command(Keyword.get(config, :cargo, :system))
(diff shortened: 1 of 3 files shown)


hansihe:
Yep, I supporting building binaries sounds like a good idea to me in general.

How does this handle multiple binary entries in the Cargo.toml file?

cristianberneanu:
> How does this handle multiple binary entries in the Cargo.toml file?

I did a quick test and it seems to work fine.

PR #168, opened 2018-09-27
Allow setting target directory

evnu:
When using workspaces, it is important that the target directory of
different NIF crates matches. Otherwise, dependencies will be compiled
multiple times. By allowing to set the target directory, we can ensure
that NIF crates are compiled to the same directory, making cargo
workspaces usable.


hansihe:
Thanks!

PR #175, opened 2019-01-04
Fix load error

### LINKED ISSUE #123: Running `mix test` does not work on the first run (needs to compile twice)
```
[Modify the `mix.exs` file]

$ mix test
Compiling NIF crate :rustler_test (native/rustler_test)...
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
Compiling 221 files (.ex)

13:15:37.820 [warn]  The on_load function for module Elixir.RustlerTest returned:
{:error, {:upgrade, 'Upgrade not supported by this NIF library.'}}

[Tests fail]

[Then, running again works]

$ mix test

[Tests succeed]
```

To make it work in travis CI I had to compile before running the tests:

```
 - MIX_ENV=test mix compile
 - mix test || travis_terminate 1
```

Same issue happens when trying to open the console with: `iex -S mix`

Rustler compiler is in the recommended order:
`compilers: [:rustler, :phoenix, :gettext] ++ Mix.compilers,`

scohen:
Fixes #123 

@scrogson FYI

### DIFF
--- a/rustler_mix/lib/rustler.ex
+++ b/rustler_mix/lib/rustler.ex
@@ -40,6 +40,10 @@ defmodule Rustler do
       @rustler_opts opts
 
       def __init__ do
+        # Remove any old modules that may be loaded so we don't get
+        # :error, {:upgrade, 'Upgrade not supported by this NIF library.'}}
+        :code.purge(__MODULE__)
+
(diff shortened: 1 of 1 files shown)


scrogson:
@scohen looks like this breaks the build.

scrogson:
:heart: Thanks for fixing this!

PR #177, opened 2019-01-14
Bump quote and syn using proc_macro2

evnu:
This pull request bumps the `quote` and `syn` crates to newer versions. As their respective API changed, changes to `rustler_codegen` were necessary (e.g. moving from `Tokens` to `TokenStream`). To simplify using `TokenStream` together with `quote` (e.g. combining multiple `TokenStream` into a single `quote!`), I added `proc_macro2`.


hansihe:
Thank you!

PR #178, opened 2019-01-17
Support enif_make_map_from_arrays

### LINKED ISSUE #157: Support enif_make_map_from_arrays
This is a new NIF function introduced in OTP 21, which is a more efficient way of building maps than repeated `enif_make_map_put` calls.

Ideally I would imagine rustler to expose a similar primitive with a fallback to the repeated `enif_make_map_put` calls solution, so that libraries can upgrade and take advantage of best performance when running on OTP 21 and have a reasonable fallback on older versions.

Reference: http://erlang.org/doc/man/erl_nif.html#enif_make_map_from_arrays

evnu:
This PR adds support for `enif_make_map_from_arrays`. This requires OTP 21.

Fix #157.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,17 @@ fn add<'a>(env: Env<'a>, args: &[Term<'a>]) -> NifResult<Term<'a>> {
 }
 ```
 
+#### Supported nif_version
+
+Rustler uses `erlang:system_info(nif_version)` to detect the supported NIF version of the Erlang/OTP
+system for which the NIF is to be compiled. It is possible to restrict the NIF version to an older
+version if the NIF is to be compiled for an older version of Erlang. For example, if the target NIF
+version should be `2.7` (Erlang/OTP 17.3), this can be defined using an environment variable:
+
+```
+RUSTLER_NIF_VERSION=2.7 mix compile
+```
+
 #### Community
 
 You can find us in `#rustler` on [freenode](http://freenode.net/) or [the elixir-lang slack](https://elixir-slackin.herokuapp.com/).
--- a/rustler/Cargo.toml
+++ b/rustler/Cargo.toml
@@ -13,6 +13,10 @@ alternative_nif_init_name = []
 erlang_nif-sys = ">=0.6.4"
 lazy_static = "1.0"
 
+[build-dependencies]
+lazy_static = "1.0"
+which = "2"
+
 [package.metadata.release]
 
 [[package.metadata.release.pre-release-replacements]]
--- a/rustler/build.rs
+++ b/rustler/build.rs
@@ -1,0 +1,61 @@
+/// Detect NIF version to build against
+
+use std::env;
+use std::process::Command;
+
+extern crate lazy_static;
+use lazy_static::lazy_static;
+
+extern crate which;
+use which::which;
+
+lazy_static! {
+    // keep this sorted by version number
+    static ref ERTS_VERSIONS: Vec<&'static str> = vec![
+        "2.7", "2.8", "2.9", "2.10", "2.11", "2.12", "2.13", "2.14"
+    ];
+}
+
+fn main() {
+    let version = match env::var("RUSTLER_NIF_VERSION") {
+        Ok(version) => version,
+        Err(_) => get_version_from_erl()
+    };
+
+    activate_versions(&version);
+}
+
+fn get_version_from_erl() -> String {
+    let erl = which("erl").expect("expected to find 'erl' executable");
+    let args = vec![
+        "-noshell",
+        "-eval",
+        r#"io:format("~s~n", [erlang:system_info(nif_version)]), init:stop()."#
+    ];
+
+    let version = Command::new(erl)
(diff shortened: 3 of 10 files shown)


evnu:
> This requires OTP 21.

This obviously fails for older versions. I could implement a fallback to repeated `map_put` calls, if there is a way to check if `enif_make_map_from_arrays` is available. Any ideas?

evnu:
> Any ideas?

One idea is to hide this behind a feature flag; alas, this makes running the tests on CI more complicated.

hansihe:
This is tricky.

I don't know if there is any better way of doing this than using feature flags. `rustler_mix` could automatically pass the feature flags for the version it's compiling for. In that case it would also make sense to make a more abstracted API for constructing maps which falls back to the old method for older versions.

What do you think?

evnu:
e472b75 adds simple detection of the ERTS version used while building rustler. This activates feature flags which can be used to select function implementations. I provided a fallback function in aeb0ad5. Lets see what CI says about this.

evnu:
@hansihe Did you find some time to think about this?

evnu:
@hansihe ping.G

hansihe:
Really sorry for being so unresponsive.

I like the base implementation of `make_map_from_arrays`, and would like to merge that immediately.

I am a bit more unsure about the way things are done with the build script. Would it be better if the `rustler_mix` passed the flags appropriate for the VM it's running in? That strikes me as a little bit more robust. On the other hand being able to build it manually is also appealing. Maybe a combination where the build script would enable features from an environment variable if it's set, and doing what it's doing here otherwise? What do you think?

evnu:
> I am a bit more unsure about the way things are done with the build script. Would it be better if the `rustler_mix` passed the flags appropriate for the VM it's running in? That strikes me as a little bit more robust.

I am not sure if this is more robust. Do you think this might be more future proof w.r.t. changes in the way OTP formats its version numbers?

Note that I am still not sure to what version (ERTS, OTP, ..) the feature flags should refer to. See https://github.com/hansihe/rustler/pull/178#issuecomment-457195898 for reference.

I just realized that `erlang_nif-sys` also uses [build.rs](https://github.com/goertzenator/erlang_nif-sys/blob/master/build.rs#L24). It uses `erlang:system_info(nif_version)`. This seems like a better way then parsing the version manually.

> On the other hand being able to build it manually is also appealing.

I use that a lot actually. My workflow is usually to run `cargo check` while implementing something in rust. This would be broken by requiring to run through `mix`. OTOH, we could add `mix rustler.check` to do the same.

> Maybe a combination where the build script would enable features from an environment variable if it's set, and doing what it's doing here otherwise?

What would that look like on the command line? `mix rustler.compile --<feature>`? Or would we need to add this into the `rustler_crates` configuration, something like `features: [:a, :b, :c]`?

hansihe:
> I am not sure if this is more robust. Do you think this might be more future proof w.r.t. changes in the way OTP formats its version numbers?

Robust as in it doesn't depend on what version of erlang is on your path to decide what to compile for, instead going by what version VM the NIF is actually going to be loaded in. Although I guess it doesn't really matter when, as you say, `erlang_nif-sys` uses whatever is on the path.

> Note that I am still not sure to what version (ERTS, OTP, ..) the feature flags should refer to. See [#178 (comment)](https://github.com/hansihe/rustler/pull/178#issuecomment-457195898) for reference.
> 
> I just realized that `erlang_nif-sys` also uses [build.rs](https://github.com/goertzenator/erlang_nif-sys/blob/master/build.rs#L24). It uses `erlang:system_info(nif_version)`. This seems like a better way then parsing the version manually.

Yeah, we should probably go by the `nif_version`.

> What would that look like on the command line? `mix rustler.compile --<feature>`? Or would we need to add this into the `rustler_crates` configuration, something like `features: [:a, :b, :c]`?

Well, the idea was that rustler_mix would set an environment variable (something like `RUSTLER_NIF_VERSION`) to the NIF version of the actual running VM. The build script then checks for the variable and uses that version if the variable is set. If it's not, then it would call the erlang that's on the path to get its NIF version. But again, I am not sure how much it matters (or if this is even wanted), if `erlang_nif-sys` just uses whatever is on the path.

More concretely, there wouldn't be any difference at all when using it normally through `rustler_mix`, nor when just running `cargo check`.

evnu:
> Robust as in it doesn't depend on what version of erlang is on your path to decide what to compile for, instead going by what version VM the NIF is actually going to be loaded in. Although I guess it doesn't really matter when, as you say, `erlang_nif-sys` uses whatever is on the path.

Ah, that makes sense! While we of course aren't able to compile for a version which is not yet supported by `erlang_nif-sys`, this would at least allow compiling for an older version. I like that idea. Should I go ahead and implement that? We can then discuss the approach further.

hansihe:
That sounds good to me!

evnu:
I implemented nif version detection and refactored the history.

hansihe:
Looks good!

PR #182, opened 2019-02-20
Add rustler attributes for conditional compilation

### LINKED ISSUE #180: Attribute to choose between encode and decode for codegen
Sometimes, I'd like to be able to derive an Encoder for a structure, but do not want (or need) a Decoder:

```rust
#[derive(NifStruct, Debug)]
struct MyStruct<'a> {
  some_item: &'a u8
}
```

Currently, `NifStruct ` cannot be derived for such a struct, as we cannot derive a Decoder for a reference (`&'a u8` in this example). It would be nice if it was possible to tell codegen that only encoding or deconding should be implemented. I imagine something like this:

```rust
#[derive(NifStruct, Debug)]
#[rustler(encode)]
struct MyStruct<'a> {
  some_item: &'a u8
}
```

evnu:
This commit adds attributes for conditional code generation of encode and
decode. Users can now choose if they want both encode and decode
functions to be derived, or only one of those functions.

For example, for the following struct `A`, only encode is derived:

```rust
struct A<'a> {
    my_string: &'a str,
}
```

Additionally, this commit also refactors how the module for Elixir
structs or the tag for Erlang records can be set. In addition to the
format `#[module = "A.B.C"]` or `#[tag = "a"]`, the `rustler` attribute
may also be used:

```rust
struct A<'a> {
    my_string: &'a str,
}
```

Fix #180

### DIFF
--- a/rustler_codegen/src/ex_struct.rs
+++ b/rustler_codegen/src/ex_struct.rs
@@ -2,25 +2,12 @@ use proc_macro2::{Span, TokenStream};
 
 use ::syn::{self, Data, Field, Ident, Meta, Lit};
 
-pub fn transcoder_decorator(ast: &syn::DeriveInput) -> TokenStream {
-    let elixir_module = {
-        let ref attr_value = ast.attrs.iter()
-            .map(|attr| attr.parse_meta())
-            .find(|meta| match meta {
-                Ok(Meta::NameValue(meta_name_value)) =>
-                    meta_name_value.ident == "module",
-                _ => false
-            })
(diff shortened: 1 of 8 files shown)


evnu:
The merge request also includes the suggestion from https://github.com/hansihe/rustler/issues/180#issuecomment-465581683. I will remove the code concerning `module` and `tag` if this is not wanted. If this suggestion should be included, I will adapt the documentation.

hansihe:
Looks good to me! Thanks!

PR #183, opened 2019-02-21
Fix link to documentation

padde:
The old link is dead. It should also be fixed in the project description on Github.

```
https://docs.rs/crate/rustler 
```


hansihe:
Thanks!

PR #188, opened 2019-02-25
Fix map_from_arrays

mischov:
Fixes problem where `enif_make_map_from_arrays` was never being used by `map_from_arrays` because it was behind an invalid `cfg` version.

- Fix cfg to be `nif_version_2_14` instead of `nif_2_14`
- Add missing lifetimes to the keys and values in the type signature

### DIFF
--- a/rustler/src/types/map.rs
+++ b/rustler/src/types/map.rs
@@ -25,8 +25,8 @@ impl<'a> Term<'a> {
     /// ```elixir
     /// List.zip(keys, values) |> Map.new()
     /// ```
-    #[cfg(nif_2_14)]
-    pub fn map_from_arrays(env: Env<'a>, keys: &[Term], values: &[Term]) -> NifResult<Term<'a>> {
+    #[cfg(nif_version_2_14)]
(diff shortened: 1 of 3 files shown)


rustler/src/types/map.rs
```diff
@@ -25,8 +25,8 @@ impl<'a> Term<'a> {
     /// ```elixir
     /// List.zip(keys, values) |> Map.new()
     /// ```
-    #[cfg(nif_2_14)]
-    pub fn map_from_arrays(env: Env<'a>, keys: &[Term], values: &[Term]) -> NifResult<Term<'a>> {
+    #[cfg(nif_version_2_14)]
+    pub fn map_from_arrays(env: Env<'a>, keys: &[Term<'a>], values: &[Term<'a>]) -> NifResult<Term<'a>> {
```
scrogson: When you fixed this to `nif_version_2_14` did it fail to compile without the lifetimes on `Term`? I suspect it should have, just curious.

mischov: It did not fail. I don't know what exactly is being tested by the test but everything passed and the debug I had in that version was called.

scrogson: Cool thanks!

PR #189, opened 2019-02-27
implement hashing for term

hansihe:


jorendorff:
Cool patches. I [asked about the salt thing](https://users.rust-lang.org/t/hashing-with-a-salt/25793) on the Rust user forums. It might be possible to do better; not sure.

hansihe:
Just to give some context as to why I am not proceeding on this.

The way conditional compilation in rustler is done right now, we depend on config flags passed directly from the build script. As far as I can tell, there is no way for depending crates to select on these flags.

This becomes a problem here, since we want to modify the tests depending on which NIF version we are on.

As far as I can tell, we have a couple of options:
1. Find a way to apply proper cargo feature flags in a way that is usable by all crates in the compilation tree. 
As far as I can see, there is no way we can do this even from a build script in the root crate. The only thing I can think of here, is making rustler_mix pass feature flags to cargo, but I don't really like this either, because that would affect `cargo check`.
2. Provide a special solution for the test suite.
That would fix this exact issue, but I don't really like this solution either, since I can definitely imagine normal users having this exact same problem.
3. Use the `nif_version_` feature flags only internally in rustler for fallback implementations, then hide every feature where we can't write a fallback implementation behind a cargo feature flag that the user explicitly has to enable from their NIF crate. As far as I can tell, this is the best option we have.

Feel free to correct me if I'm wrong on any points, my understanding of how cargo features/cfg flags works is a bit shaky.

evnu:
As we do not support `nif_version < 2.14` any longer (https://github.com/rusterlium/rustler/pull/387), work here can proceed I think. No build flags required, rebase should be sufficient.

hansihe:
This was merged in https://github.com/rusterlium/rustler/pull/410

PR #191, opened 2019-03-05
Add documentation for compiling binaries on OSX

KevinCybura:
When attempting to Link on OSX ive been getting linker errors similar to this Issue 
https://github.com/hansihe/rustler/issues/151
I've done some investigation and saw that there was an attempt to solve this problem here
https://github.com/hansihe/rustler/pull/174
However it didn't solve it for me until I add the .cargo/config to my project. So I think add this documentation would be useful

PR #192, opened 2019-03-06
Generate .cargo/config

scrogson:
This PR adds a `.cargo/config` to the generated skeleton to avoid issues
with running `cargo build` on macOS.

I've added a warning to the compiler task if compiling on macOS and the
`.cargo/config` does not exist.

Here is an example warning when the file doesn't exist:

```
λ mix compile
==> rustler
Compiling 1 file (.ex)
==> hyper
Compiling NIF crate :hyperbeam (native/hyperbeam)...

Compiling on macOS requires special link args in order to compile
correctly.

Rustler is currently working around this issue in the compiler task.
This will be removed in v1.0.0 in favor of a user supplied .cargo/config
file.

To remove this warning, please create /Users/scrogson/github/rusterlium/hyper/native/hyperbeam/.cargo/config
with the following content:

      [target.x86_64-apple-darwin]
      rustflags = [
          "-C", "link-arg=-undefined",
          "-C", "link-arg=dynamic_lookup",
      ]

See https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/MachOTopics/1-Articles/executing_files.html
for more details.


    Finished release [optimized] target(s) in 0.15s
```

Fixes #191


evnu:
Does the config need to be replicated for all crates in a workspace, or is a single configuration per workspace enough?

scrogson:
> Does the config need to be replicated for all crates in a workspace, or is a single configuration per workspace enough?

@evnu I believe a single `.cargo/config` will work at the root of the workspace (as seen in the `rustler` repository itself).

evnu:
>  I believe a single .cargo/config will work at the root of the workspace (as seen in the rustler repository itself).

Should the test here than traverse the directories to find any matching config?

scrogson:
> Should the test here than traverse the directories to find any matching config?

Not sure what you mean. It appears that a top-level `.cargo/config` would suffice.

Do you have a workspace-based NIF project you could test this hypothesis with?

evnu:
> Not sure what you mean. It appears that a top-level .cargo/config would suffice.

I think that the test for the configuration file [here](https://github.com/hansihe/rustler/pull/192/files#diff-5a3599e65b63b9eed8eb2275fae46987R100) might output an "incorrect" warning, if the `.cargo/config` is not at the exact place where it is expected. This doesn't hurt, but might be confusing at first.

PR #201, opened 2019-04-09
Add proc_key/1 function match for literals

erik-nymi:
Supports Cargo.toml configurations such as:
```toml
[target.'cfg(windows)'.dependencies]
```
which are matched by the lexer as `{:literal, _}` but fail to parse with a function missing error.


scrogson:
@erik-nymi looks like this PR was based on the `vendor-erl-nif-sys` branch?

Can you please close this and base your changes on `master` instead?

PR #207, opened 2019-05-04
Add support for decoding binaries from iolists

mpaltun:


hansihe:
I like this in general, thanks!

Only thing is, I think it might be a better idea to not actually include it in the Decoder for Binary, and instead have a separate function on Term that does this. I think having the Binary Decoder do this implicitly might not be what most people expect to happen.

mpaltun:
Right. That's was something that I wasn't really sure. 

I updated the code as you suggested. Please, let me know if you have any other suggestions.

hansihe:
Thanks!

PR #213, opened 2019-05-24
Re-export codegen derives

scrogson:
This change removes the need for users to depend directly on
`rustler_codegen`.


scrogson:
@hansihe should we make this into a feature instead? Perhaps make it default? Or just leave as is?

hansihe:
This looks good, but I agree it might be a good idea to make it a feature and mark it as defaul

PR #217, opened 2019-05-24
Use toml-elixir over self-rolled parser

davydog187:
[toml-elixir](https://github.com/bitwalker/toml-elixir) is was built for Distillery and is better tested and fully featured. This is a small change, but it helps take noise out of `rustler_mix`.

Note that this changes the minimum supported Elixir version to be `1.6` due to https://github.com/bitwalker/toml-elixir/blob/master/mix.exs#L11


davydog187:
@scrogson before we merge, do you have a suggested way to test the `mix` task?

davydog187:
merge?

PR #218, opened 2019-05-25
Add .formatter.exs, check formatting in Travis

davydog187:
The problem with adding the formatter to Travis is that it will always fail on Elixir < 1.6, and since the formatter outputted slightly different on Elixir version 1.6, that will fail too.

In #217 I dropped support for Elixir < 1.6. 

Elixir 1.6 was released `January 17, 2018`
Elixir 1.7 was released `July 25, 2018`

I'm not sure what Rustler's minimum support policy is, but happy to accommodate to make it more flexible.


scrogson:
Closing in favor of #253

Issue #215, opened 2019-05-24
UntaggedEnum with bool variant breaks unexpectedly


davydog187:
As shown by #214 , if you have an UntaggedEnum with a bool variant, it will break the decoding of struct variants.

```rust
#[derive(NifUntaggedEnum)]
pub enum UntaggedEnum {
    Foo(u32),
    Bar(String),
    Baz(AddStruct),
    Bool(bool), // Only addition
}
```

```bash
  1) test untagged enum transcoder (RustlerTest.CodegenTest)
     test/codegen_test.exs:44
     Assertion with == failed
     code:  assert :invalid_variant == RustlerTest.untagged_enum_echo([1, 2, 3, 4])
     left:  :invalid_variant
     right: true
     stacktrace:
       test/codegen_test.exs:48: (test)
```

I have started looking at `rustler_codegen/src/untagged_enum.rs`, but nothing stuck out to me that would cause this issue, and I'm wondering if this is an issue that needs to be fixed upstream in `syn`.

I would love to help tackle this bug, but looking for guidance on where I should start poking. Thank you

evnu:
The decoder for `bool` uses [truthiness](https://github.com/rusterlium/rustler/blob/master/rustler/src/types/primitive.rs#L57) to determine if the bool is true or false:

```rust
impl<'a> Decoder<'a> for bool {
    fn decode(term: Term<'a>) -> NifResult<bool> {
        Ok(atom::is_truthy(term))
    }
}
```

[is_truthy](https://github.com/rusterlium/rustler/blob/master/rustler/src/types/atom.rs#L121) checks if the term does not equal `:false` or `:nil`:

```rust
pub fn is_truthy(term: Term) -> bool {
    !((term.as_c_arg() == false_().as_c_arg()) || (term.as_c_arg() == nil().as_c_arg()))
}
```

I too find this behaviour surprising and think it should be made explicit. Where truthiness is expected, an newtype `Truthy` could be used:

```rust
struct Truthy(bool);
```

evnu:
Can this be closed as #219 is now merged, or is there more to do here?

davydog187:
Closing!

PR #214, opened 2019-05-24
UntaggedEnum with bool variant breaks unexpectedly

davydog187:
This PR does not have a fix yet, but opening for a discussion

```bash
  1) test untagged enum transcoder (RustlerTest.CodegenTest)
     test/codegen_test.exs:44
     Assertion with == failed
     code:  assert :invalid_variant == RustlerTest.untagged_enum_echo([1, 2, 3, 4])
     left:  :invalid_variant
     right: true
     stacktrace:
       test/codegen_test.exs:48: (test)
```

PR #219, opened 2019-05-28
Introduce a Truthy newtype

evnu:
This pull request implements the `Truthy` newtype suggested in https://github.com/rusterlium/rustler/issues/215#issuecomment-496675299. This type allows to distinguish between "real" booleans and truthy values (terms not equal `false` or `nil`). Where a "real" boolean is expected, Rust's `bool` can be used directly. Where a truthy value is expected, the newtype `Truthy` can be used.

See #214 for the original failing test case which is fixed by this pull request.

### DIFF
--- a/rustler/src/types/truthy.rs
+++ b/rustler/src/types/truthy.rs
@@ -1,0 +1,24 @@
+//!
+//! A type to represent a truthy value.
+//!
+//! In Elixir, a term which does not equal `:false` or `:nil` is considered to be
+//! truthy. This does not cleanly map to Rust's `bool` type. To distinguish between
+//! `bool` and a truthy value, the newtype `Truthy` can be used.
+//!
+
+use crate::types::atom;
+use crate::{Decoder, Encoder, Env, NifResult, Term};
+
+pub struct Truthy(bool);
+
+impl Encoder for Truthy {
+    fn encode<'a>(&self, env: Env<'a>) -> Term<'a> {
+        self.0.encode(env)
+    }
+}
+
+impl<'a> Decoder<'a> for Truthy {
+    fn decode(term: Term<'a>) -> NifResult<Truthy> {
+        Ok(Truthy(atom::is_truthy(term)))
+    }
+}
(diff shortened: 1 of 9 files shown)


rustler/src/types/truthy.rs
```diff
@@ -0,0 +1,24 @@
...
+//!
+//! In Elixir, a term which does not equal `:false` or `:nil` is considered to be
+//! truthy. This does not cleanly map to Rust's `bool` type. To distinguish between
+//! `bool` and a truthy value, the newtype `Truthy` can be used.
+//!
+
+use crate::types::atom;
+use crate::{Decoder, Encoder, Env, NifResult, Term};
+
+pub struct Truthy(bool);
```
davydog187: I really dislike adding this additional layer of abstraction. The point is not to model Elixir directly, but to be able to map directly to Rust types.

I would expect that if I tried to decode an Elixir value as `bool` that is not `:true` or `:false`, it would fail.

This is an awkward abstraction that does not provide the intent I'm looking for

davydog187: Maybe I'm misunderstanding this. Is this just to provide the option to decode `Truthy` values, but not mandate them?

Looking at the test code, it looks like you're adding a feature and fixing the bug in #214

hansihe:
I like this in general, but I am concerned about backwards compatibility.

I think we should at the very least we should start making a proper changelog/release notes. I guess a CHANGELOG.md in the repo root would do the trick?

davydog187:
That would be really easy to do going forward. Semantic versioning helps too!

evnu:
> I guess a CHANGELOG.md in the repo root would do the trick?

With the increase in popularity of rustler, a changelog would come in handy. It appears that the upcoming release will have multiple things that a user should take care of (e.g. deciding to drop the explicit `rustler_codegen` dependency). @hansihe Do you want to start a changelog? I would then rebase the pull request and add a section on the changes here.

hansihe:
@evnu I added a Changelog to the repo

evnu:
@hansihe I added a short entry. Do you think that is sufficient?

scrogson:
Maintaining the changelog from each PR is usually not fun.

There ends up being tons of merge conflicts after each PR is merged. This ends up blocking all the rest of the PRs until the author can find the time to fix it up.

scrogson:
I guess now that we can fix up simple conflicts in the UI it won't be that big of a deal breaker.

@hansihe we should probably create a CONTRIBUTING.md and a PR template...

evnu:
@hansihe Can I rebase on top of master to fix the clippy warnings, or do you prefer merges?

hansihe:
@evnu Rebasing is fine with me!

evnu:
Travis was [stuck](https://travis-ci.org/rusterlium/rustler/jobs/544398022). I pushed again to trigger a new build.

PR #220, opened 2019-06-03
Better struct decoding error messages

davydog187:
Currently if you try to decode a struct that has a field with a type mismatch, all you get back is `Argument error`.

We use Rustler to build large application level NIFs, with structs that have many, often nested, fields. This PR attempts to improve the error message to at least pinpoint which field cannot be decoded. I would love feedback on how I could further improve. 

1. Is there a way to raise an Error that is not `ErlangError`?
2. The error message currently now shows the offending field and the struct `Could not decode field :lhs on %AddStruct{}`, should it also show the value?

### DIFF
--- a/rustler_codegen/src/ex_struct.rs
+++ b/rustler_codegen/src/ex_struct.rs
@@ -26,6 +26,7 @@ pub fn transcoder_decorator(ast: &syn::DeriveInput) -> TokenStream {
         .map(|field| {
             let ident = field.ident.as_ref().unwrap();
             let ident_str = ident.to_string();
+
             let atom_fun = Ident::new(&format!("atom_{}", ident_str), Span::call_site());
 
             quote! {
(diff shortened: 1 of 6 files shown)


rustler_codegen/src/ex_struct.rs
```diff
@@ -82,8 +83,12 @@ pub fn gen_decoder(
             let ident = field.ident.as_ref().unwrap();
             let ident_str = ident.to_string();
             let atom_fun = Ident::new(&format!("atom_{}", ident_str), Span::call_site());
+            let error_message = format!("Could not decode field :{} on %{}{{}}", ident.to_string(), struct_name.to_string());
```
davydog187: This should probably be moved to inside the error case to avoid any penalty performance penalties

evnu:
I think this would be a useful addition for all struct-like types (NifStruct, NifRecord, NifMap), as well as for `NifTuple`.

davydog187:
I agree! I would love some feedback on my questions and then I can add to the other types as well

rustler_codegen/src/ex_struct.rs
davydog187: Resolving, this happens at compile time, so we don't care! 😄

hansihe:
This looks good to me!

If you want to improve the error messages for the other ones as well, that would be great.

It would be nice if you could add the changes to the (newly added) changelog as well.

davydog187:
@hansihe I added additional error messages for map, record and tuple. There's much more I can do in the way of error messages for other types, but I think that could be in different pull requests as I'm not sure how deep that should go at the moment.

hansihe:
Looks good to me, thanks!

PR #223, opened 2019-06-05
Add clippy

### LINKED ISSUE #222: Make use of clippy?
I played around with `cargo clippy` and rustler the other day and wondered if there is interest to bring that upstream. I fixed the warnings from `cargo clippy --all-features --all-targets` in a [branch](https://github.com/evnu/rustler/tree/clippy). The suggestions mainly aimed at removing superfluous (as of Rust 2018) lifetime annotations, removing unnecessary captures and replacing references where move is more appropriate due to `Copy` being implemented for a type. But clippy found two things which might be of interest:

1. `bin.as_mut_slice().write(self.as_bytes()).expect("memory copy of string failed")` should use `write_all()` instead of `write`, as `write` does not guarantee to write all at once
2. `ptr.offset(offset as isize)` should be `ptr.add(offset)`, as `offset` can never be a negative number.

evnu:
This pull requests adds [clippy](https://github.com/rust-lang/rust-clippy) to travis. I fixed the current clippy suggestions within Rustler. I left out some suggestions to show what they look like.

To run clippy against Rustler:

```
$ rustup component add clippy
$ cargo clippy --all-targets --all-features -- -D warnings
```

Fix #222.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,9 +23,12 @@ before_install:
 install:
   - mix local.rebar --force
   - mix local.hex --force
+  - source $HOME/.cargo/env
+  - rustup component add clippy
 
 script:
   - source $HOME/.cargo/env
   - cargo test
   - cd rustler_mix && mix do deps.get, test && cd -
   - cd rustler_tests && mix do deps.get, test && cd -
+  - cargo clippy --all-targets --all-features -- -D warnings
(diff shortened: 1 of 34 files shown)


evnu:
The remaining errors are in `erlang_nif_sys`. I will fix them if there is interest to merge this pull request.

hansihe:
Definitely interested in merging this.

Would be nice if you could fix the remaining errors and add the change to the changelog

evnu:
> Would be nice if you could fix the remaining errors and add the change to the changelog

Will do.

evnu:
> Will do.

Done.

PR #228, opened 2019-07-17
Specifies so path to @external_resource

Shikanime:
## What kind of change does this PR introduce?

**Development experience improvement**.

## What is the current behaviour?
When using IEX when we use `recompile/0', the nif is not reloaded.

## What is the new behaviour?
We explicitly specify that the module depends on an external SO file.

## Does this PR introduce a breaking change?

**No**.

### DIFF
--- a/rustler_mix/lib/rustler.ex
+++ b/rustler_mix/lib/rustler.ex
@@ -81,12 +81,12 @@ defmodule Rustler do
 
     lib_path = Application.app_dir(otp_app, "priv/native/#{crate}") |> to_charlist()
     load_data = opts[:load_data] || config[:load_data] || 0
-    resource_path = Path.join(lib_path, ressource_extension())
+    resource_path = Path.join(lib_path, resource_extension())
 
     {resource_path, lib_path, load_data}
   end
 
-  defp ressource_extension do
+  defp resource_extension do
     case :os.type() do
       {:win32, _} -> ".dll"
       {:unix, _} -> ".so"
--- a/rustler_mix/lib/rustler.ex
+++ b/rustler_mix/lib/rustler.ex
@@ -35,17 +35,19 @@ defmodule Rustler do
 
   defmacro __using__(opts) do
     quote bind_quoted: [opts: opts] do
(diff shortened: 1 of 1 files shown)


NobbZ:
This will lead to recompiling the module when not necessary…

The proper way is to re-load the module rather than recompiling it in iex.

rustler_mix/lib/rustler.ex
```diff
@@ -77,10 +79,17 @@ defmodule Rustler do
...
+    lib_path = Application.app_dir(otp_app, "priv/native/#{crate}") |> to_charlist()
     load_data = opts[:load_data] || config[:load_data] || 0
-    so_path = String.to_charlist("#{priv_dir}/native/#{crate}")
+    resource_path = Path.join(lib_path, ressource_extension())
 
-    {so_path, load_data}
+    {resource_path, lib_path, load_data}
+  end
+
+  defp ressource_extension do
```
scrogson: I think you meant `resource_extension`?

Shikanime:
@NobbZ I see what you mean but as a module that host a Nif is usually very light, I don't see any big  problem. Specifying the So/DLL to an external resource will help to trigger the `@on_load` annotation automatically, considering that the developper should not be aware of the internal `__init__` function.

From the developer's point of view, the fact that the `mix compile' command compiles Erlang, Elixir and Rust code. We should expect kind of same behaviour when using the `recompile` command.

NobbZ:
Well, its not even given, that the `so`/`dll` has even changed, unless you compiled manually before that. Since you already have recompiled manually, just reloading should be enough.

Also it just feels semantically wrong.

Although I admit, you are totally correct on "they are usually small", but we never know how other modules might compile-time depend on those NIFs, which might trigger much more recompilation (but probably in a good way as well).

So at the end, yes, I think the change is worth it.

Shikanime:
Oh yes, you're right, it seems a little semantically wrong.
So to continue debugging without leaving IEX, we should just call the `MyNif.__init__()` to manually reload the nif.

PR #229, opened 2019-07-28
Fix segfault in map_from_arrays.

jeremyjh:
Currently on master if you pass a shorter list of values than keys to `Term.map_from_arrays` you will get a segfault (you can reproduce with the new test case before the patch is applied). This adds a check that the sizes are the same and returns BadArg otherwise.


Shnatsel:
This looks like out-of-bounds access (i.e. buffer overflow), which could be an exploitable security vulnerability. Please consider filing a security advisory at https://github.com/RustSec/advisory-db so that users can check if they're using a vulnerable version and upgrade.

Issue #107, opened 2017-07-12
Exposing enif_schedule_nif safely


jorendorff:
`enif_schedule_nif` comes with a contract, much like `enif_raise_exception`. It's hard to support safely.

But we already support `enif_raise_exception` (see https://github.com/hansihe/rustler/blob/master/src/codegen_runtime.rs#L44 ). We could support `enif_schedule_nif` in the same way. Don't let the user call it directly, but let them return a special Rust value that means "call `enif_schedule_nif` safely for me".

There is one problem that we would need to solve, though. Currently, all NIFs have the return type `NifResult<NifTerm<'a>>`, allowing just two possibilities:

* `Ok(term)`: "return this term"
* `Err(exc)`: "raise this exception"

Now we want to add a third option, "schedule this nif". But we can't just change the return type, because that would break all existing code. I guess the way to do this is with a trait. Come to think of it, we should allow a wide variety of return types anyway, not just `NifResult<NifTerm<'a>>`; for example, anything that implements `NifEncode` should be just as good as a `NifTerm`, right?

treyzania:
Would it really be a bad thing to change the return type to add a `Sched(...)` option?  It obviously would break existing code trying to upgrade, but Rustler isn't at the magical "1.0" yet so there *is* a little more flexibility.

jorendorff:
I think using a trait is better. The idea is that the *user* decides what the return type is, and Rustler figures out what to do with it.

So users may keep using `NifResult<NifTerm>` ... and write a new NIF that returns `NifResult<NifSched>` ... and write another new NIF that returns a Rust `String`, there's no reason Rustler couldn't handle all of them.

jorendorff:
I don't know how readable this is, but here's the approach I took for a little Scheme implementation I've been messing around with: https://github.com/jorendorff/cell-gc/commit/149ce249bb4ac565783deeef9bcdbc502d061e5d

Before:

```rust
fn char_upcase<'h>(_hs: &mut GcHeapSession<'h>, args: Vec<Value<'h>>) -> Result<Trampoline<'h>> {
    if args.len() != 1 {
        return Err("char-upcase: 1 argument required".into());
    }
    let c = args[0].clone().as_char("char-upcase: character required")?;

    // I think the only character that doesn't upcase to a single character is
    // U+00DF LATIN SMALL LETTER SHARP S ('ß'). Guile returns it unchanged.
    // Fine with me.
    let mut up = c.to_uppercase();
[...]

Issue #106, opened 2017-07-01
Exposing `enif_schedule_nif`


Qqwy:
As stated in [issue #44's discussion](https://github.com/hansihe/Rustler/issues/44#issuecomment-269916075), `enif_schedule_nif` is currently not exposed by Rustler because it is not known how to expose the function in a compile-time safe way,

as the NIF is supposed to return to the Erlang scheduler right after `enif_schedule_nif` was called.

My proposal would be to expose `enif_schedule_nif` in an `unsafe` context, to ensure that people that want to use it are mindful of if they are using it in the proper way.

In the future, it might be possible to wrap `enif_schedule_nif` in some way.
An example would be as a preemptive Iterator-consumer, to ensure that the NIF will be split in chunks of `O(1)`, even when performing `O(n)` or `O(n^2)` etc iterations.

hansihe:
I totally agree with this, if there is an API we can't expose safely, we should do it unsafely with warnings.

If you want to give this a go, I would be fully open to pull requests. I might give it a go next week after I am done with my talk at PolyConf.

hansihe:
Closing this issue in favour of https://github.com/hansihe/rustler/issues/107, it contains more discussion

PR #232, opened 2019-08-24
schedule_nif improvements

sunny-g:
Re-implemented for the new `#[rustler::nif]` API (excellent work btw).

Let me know what y'all think! Also open to renaming things if y'all think they make more sense. 

Previous discussion: https://github.com/rusterlium/rustler/issues/106 https://github.com/rusterlium/rustler/issues/107

### DIFF
--- a/rustler/src/lib.rs
+++ b/rustler/src/lib.rs
@@ -49,7 +49,7 @@ pub mod dynamic;
 pub use crate::dynamic::TermType;
 
 pub mod schedule;
-pub use crate::schedule::SchedulerFlags;
+pub use crate::schedule::{Schedule, SchedulerFlags};
 pub mod env;
 pub use crate::env::{Env, OwnedEnv};
 pub mod thread;
--- a/rustler/src/schedule.rs
+++ b/rustler/src/schedule.rs
@@ -49,6 +49,7 @@ pub fn consume_timeslice(env: Env, percent: i32) -> bool {
 ///         Schedule::Result(result)
 ///     } else {
 ///         // alternatively `Schedule::Next2(std::marker::PhantomData, input - 1, result * input)`
+///         // alternatively `(input - 1, result * input).into()`
 ///         Schedule::next2(input - 1, result * input)
 ///     }
 /// }
--- a/rustler/src/schedule.rs
+++ b/rustler/src/schedule.rs
@@ -1,3 +1,6 @@
+use rustler_sys::c_char;
+
+use crate::codegen_runtime::{NifReturnable, NifReturned};
 use crate::wrapper::ErlNifTaskFlags;
 use crate::Env;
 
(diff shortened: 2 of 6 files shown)


sunny-g:
@scrogson @filmor

scrogson:
Hey @sunny-g! Thanks for updating this.

I've got some ideas here that I want to share but I don't have the time to articulate it thoroughly right now. I'll take some time over the next couple days to put a review together.

sunny-g:
@scrogson fantastic, looking forward to them!

scrogson:
@sunny-g here's what I'm thinking...

Simplify the `Schedule` enum to this...

```rust
pub enum Schedule<T> {
    Return(T),
    Continue(Nif, Vec<Term>),
}

impl<T> NifReturnable for Schedule<T> where T: NifReturnable {
    ...
}
```

Then create a macro which builds the correct data structure:

```rust
reschedule!(my_nif, a, b, c, ...);
```

sunny-g:
@scrogson As you might remember, I had that in the original PR.

I changed it because the original solution required your NIF to be `Env`-aware, which was fine with the original rustler macros since all NIFs had to take in an `Env`, but seemed messier than necessary when the new macros would do the encoding/decoding to/from `Term`s for you.

Granted, a pseudo-variadic `enum` isn't really "clean" either, but that's why it implements `From<(..)>`/`Into<(..)>` with default type parameters, so you only need to specify the types and args you use.

I think regardless I should change `Schedule::Result` to `Schedule::Return` and the others to `Continue`.

Let me know if this explanation changes anything - if not at all, I'll go with what you suggested. Alternatively, I could split the difference and add a similar macro to what you've suggested that's more ergonomically variadic.

scrogson:
> As you might remember, I had that in the original PR.

I went back and looked at the original commits as I didn't remember what it looked like before, and yes it is quite similar to my suggestion.

> I changed it because the original solution required your NIF to be `Env`-aware, which was fine with the original rustler macros since all NIFs had to take in an `Env`, but seemed messier than necessary when the new macros would do the encoding/decoding to/from `Term`s for you.

With the new syntax, some things still need an `Env` anyways, so I don't see any issue with this solution requiring `Env` in order to work (which reminds me...my `reschedule!` macro sketch was missing the `Env` and scheduler flag).

scrogson:
Another thought here is maybe we only need a macro or some function to convert the args? And use:

```rust
let args: Vec<Term> = collect_terms!(env, a, b, c, ...);
Schedule::Continue(my_nif, args, DirtyCpu)
```

scrogson:
```rust
#[rustler::nif]
fn scheduled_fac(env: Env, input: u32, result: Option<u32>) -> Schedule<u32> {
    let result = result.unwrap_or(1);
    if input == 0 {
        Schedule::Result(result)
    } else {
        Schedule::Continue(schedule_fac, args!(env, input - 1, result * input), DirtyCpu)
    }
}
```

sunny-g:
2 things:

1. the last comment looks fine, but one thing to note: are the flags required, or are you specifying it to just highlight that they might differ from the NIF macro definition? If so, the enum would still need work.
2. can you help me understand the issue you're having with the PR as it is currently? Ergonomics, maintainability, extensibility, or something else? I was attempting something that in the best case would be the minimally unintuitive, and in the worst case, would be verbose but still rust-y.

sunny-g:
> The schedule flag should be allowed to override the original NIF when needed. I personally haven't found a direct need for this, but I have heard of some instances where depending on the input length you may not need to run something on the dirty schedulers and not pay the extra cost.

Makes sense.
 
> Yes, I think we just add the schedule flag in the 3rd position in the `Continue` variant.

I meant in the sense that it couldn't be "optional", so the alternative would requiring that you pass a flag, even if it's normal, which maybe isn't that big of a deal.
 
> The 2 things that I'm not happy with are the return type and the various functions to represent different arity. It feels a bit unintuitive to me.

That's understandable, but is also the reason I added the `Into` conversion, so that you wouldn't have to worry about it at all.
 
> I'm now wondering if an elegant solution could be to generate a `reschedule` function at compile-time for each NIF?
> 
> ```rust
> scheduled_fac::reschedule(env, input - 1, Some(result * input), DirtyCpu)
> ```
> 
> This would probably allow us to ensure type safety on what's passed in (which I think is part of the why behind your current implementation?)

This is actually the approach I started with on this pass, but had problems with it because inside the function, the `Ident` of the function's name is scoped to the function itself, not the NIF type that gets created.

I'll try again anyways, and regardless will also add a macro so that you don't have to use any variants.

sunny-g:
@scrogson so I wasn't able to work around the scoping issue, but have now added a `reschedule` macro that I think gets us the best of both worlds in terms of terseness and ergonomics:

```rust
#[rustler::nif]
fn scheduled_fac<'a>(input: u32, result: Option<u32>) -> Schedule<scheduled_fac, u32, u32, u32> {
    let result = result.unwrap_or(1);
    if input == 0 {
        Schedule::Return(result)
    } else {
        reschedule!(SchedulerFlags::Normal, input - 1, result * input)
    }
}
```

evnu:
Can we go forward with this PR? I think convenient rescheduling would be very nice to have. What is missing to bring this forward?

sunny-g:
@evnu Lemme take another look at this, I might be able to improve the API a bit. Otherwise it works.

dimitarvp:
Sorry to bother -- does anyone has the time to tackle this? It's a nice feature.

dimitarvp:
25 months later -- what's the blocker here? Still sounds like an excellent feature.

PR #238, opened 2019-09-18
Fall back to precompiled API snippet if gen_api.erl fails

### LINKED ISSUE #236: docs.rs broken
(I guess) since rustler_sys (erlang_nif-sys, erl_nif_sys) is now part of the normal rustler build process, the docs.rs build fails: https://docs.rs/crate/rustler/0.21.0/builds/183891

This means there is no nice way to check rustler's docs for the current version online.

filmor:
Should fix #236 and similar issues.

### DIFF
--- a/rustler/build.rs
+++ b/rustler/build.rs
@@ -16,16 +16,15 @@ lazy_static! {
 }
 
 fn main() {
-    let version = match env::var("RUSTLER_NIF_VERSION") {
-        Ok(version) => version,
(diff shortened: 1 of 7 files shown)


filmor:
Should be rebased on #233.

filmor:
@scrogson @hansihe Do you need help in the maintenance? There are quite a few PRs pending and in particular the brokenness of docs.rs right now makes it a bit tricky to use the library.

PR #240, opened 2019-09-20
Remove superfluous use statements from proc-macros

evnu:
This pull request removes some superfluous `use` statements within `rustler_codegen`.


scrogson:
Thank you!

PR #242, opened 2019-10-06
Fix a Clippy warning (Clippy 0.0.212 / Rust 1.38.0)

tatsuya6502:
Clippy in the latest Rust stable emits the following warning:

```console
error: You checked before that `unwrap()` cannot fail. Instead of checking and unwrapping, it's better to use `if let` or `match`.
  --> rustler/src/resource.rs:90:18
   |
88 |     if res.is_some() {
   |        ------------- the check is happening here
89 |         Some(ResourceType {
90 |             res: res.unwrap(),
   |                  ^^^^^^^^^^^^
   |
   = note: `-D clippy::unnecessary-unwrap` implied by `-D warnings`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_unwrap
```
This pull request addresses the warning by changing the codes as [suggested by Clippy](https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_unwrap).

**Version Info**

Rust 1.38.0 stable

```console
$ rustc -V
rustc 1.38.0 (625451e37 2019-09-23)

$ cargo clippy -V
clippy 0.0.212 (3aea860 2019-09-03)
```

PR #249, opened 2019-10-16
Rename macros

scrogson:
Renames the following macros:

* `rustler_export_nifs!` => `init!`
* `rustler_atoms!` => `atoms!`
* `resource_struct_init!` => `resource!`

I have also made the old macros still available and marked them as
`#[deprecated]`

Here are the deprecation warnings printed when using the old macros:

```
warning: use of deprecated item 'rustler::rustler_atoms': Please use `atoms!` instead.
 --> rustler_tests/native/rustler_test/src/test_atom.rs:5:5
  |
5 |     rustler::rustler_atoms! {
  |     ^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(deprecated)]` on by default

warning: use of deprecated item 'rustler::resource_struct_init': Please use `resource!` instead.
  --> rustler_tests/native/rustler_test/src/test_resource.rs:16:5
   |
16 |     rustler::resource_struct_init!(TestResource, env);
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: use of deprecated item 'rustler::rustler_export_nifs': Please use `init!` instead.
  --> rustler_tests/native/rustler_test/src/lib.rs:17:1
   |
17 | rustler::rustler_export_nifs!(
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

**EDIT**: I've added a new test crate (`deprecated_macros`) that ensures that the changes in the macros haven't broken the existing ones.


filmor:
How about we use the renaming to also improve the macros a bit? I never really understood, why atoms would be prefixed, `atoms!{ a, b, c ="d" }` would work as well.

rustler/src/export.rs
```diff
@@ -15,15 +15,27 @@
 /// The third argument is an `Option<fn(env: &Env, load_info: Term) -> bool>`. If this is
 /// `Some`, the function will execute when the NIF is first loaded by the BEAM.
 #[macro_export]
+#[deprecated(since = "0.22.0", note = "Please use `init!` instead.")]
 macro_rules! rustler_export_nifs {
+    ($name:expr, [$( $exported_nif:tt ),+,], $on_load:expr) => {
```
evnu: I am curious, should'nt  the second branch  with `*`already cover this as well?

scrogson: Yeah, that seems to work. :+1:

evnu:
Does it make sense to keep `rustler_atoms! ` if the old Syntax for declaring asks does not work anymore?

scrogson:
> Does it make sense to keep `rustler_atoms! ` if the old Syntax for declaring asks does not work anymore?

@evnu the `rustler_atoms!` version does still work (see cd64245)

scrogson:
> How about using , instead of ; as well?

Oooh...yeah, I had that thought earlier. 👍

scrogson:
Unfortunately, adding the use of the deprecated macro fails the build. How do y'all feel about adding `-A deprecated` to the `cargo clippy` check?

CHANGELOG.md
```diff
@@ -7,7 +7,34 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Dependencies have been updated.
+- Derive macros have been refactored.
+- Macros have been renamed and old ones have been deprecated:
+  - `rustler_export_nifs!` is now `rustler::init!`
+  - `rustler_atoms!` is now `rustler::atoms!`
+  - `resource_struct_init!` is now `rustler::resource!`
+- New `rustler::atoms!` macro removed the `atom` prefix from the name:
+
+```rs
+# Before
```
evnu: I guess this should be `//` comments.

scrogson: Ah, yup.

rustler/src/types/atom.rs
```diff
@@ -197,6 +194,44 @@ unsafe impl Send for Atom {}
...
+        )*
+    };
+
+    // Internal helper macros.
+    { @internal_make_atom($env:ident, $name:ident) } => {
+        $crate::atoms!(@internal_make_atom($env, $name = stringify!($name)))
+    };
+    { @internal_make_atom($env:ident, $name:ident = $str:expr) } => {
+        $crate::types::atom::Atom::from_str($env, $str)
+            .ok().expect("rustler::atoms!: bad atom string")
```
evnu: I think expect() should be enough here. Atom::from_str() returns NifResult<T>, which is just Result<T, Error>, and Result implements expect().

evnu: See the corresponding [clippy lint](https://rust-lang.github.io/rust-clippy/master/#ok_expect). I think clippy cannot find this here as it is hidden in a macro.

scrogson: Yeah, I think I saw clippy complain about that in another PR...not sure why it's not consistently erroring on this for everything?

evnu: I guess it cannot determine in this case that `Atom::from_str()` returns a NifResult because this is part of a macro and prefixed with `$crate`. I actually don't know if clippy can check macro_rules code at all.

scrogson: It looks like we can't change this here because `Error` doesn't implement `Debug`. We can make that change in #242 

```
error[E0599]: no method named `expect` found for type `std::result::Result<types::atom::Atom, error::Error>` in the current scope
   --> rustler/src/types/atom.rs:229:14
    |
229 |               .expect("rustler::atoms!: bad atom string")
    |                ^^^^^^
...
271 | / atoms! {
272 | |     /// The `nil` atom.
273 | |     nil,
274 | |
...   |
304 | |     last,
305 | | }
    | |_- in this macro invocation
    |
    = note: the method `expect` exists but the following trait bounds were not satisfied:
            `error::Error : std::fmt::Debug`
```

evnu: :+1:

filmor: @scrogson Did you mean #248?

rustler/src/types/atom.rs
```diff
@@ -197,6 +194,44 @@ unsafe impl Send for Atom {}
 /// The only overhead is checking that the atoms have been created (an atomic integer read).
 ///
 #[macro_export]
+macro_rules! atoms {
+    {
+        $(
+            $( #[$attr:meta] )*
+            $name:ident $( = $str:expr )*,
```
filmor: In newer Rust versions, you can use `?` instead of `*`, which also disallows `bla = "a" = "b"`.

filmor: Also, I think the comma should be optional, in line with other usages, so `atoms!{ ok }` should work. To do this, it should be enough to change the rule to
```rust
        $(
            $( #[$attr:meta] )*
            $name:ident $( = $str:expr )?
         ),* $(,)?
```

(see https://stackoverflow.com/questions/43143327/how-to-allow-optional-trailing-commas-in-macros/43143459#43143459)

rustler/src/types/atom.rs
scrogson: Yep, that one 😄

rustler/src/types/atom.rs
scrogson: Neat.

scrogson: Done.

Issue #241, opened 2019-09-24
Field atoms are duplicated in proc-macro


evnu:
Currently, the atoms generated for accessing fields in structs within `ex_struct.rs`, `record.rs`, etc., are duplicated for both encode and decode. This increases the amount of code generated considerably for larger structs. For example, when using [`cargo expand`](https://crates.io/crates/cargo-expand) on https://github.com/evnu/NifIo/tree/macos-platform-hack, about 200 lines of code are generated for the atoms alone, which is 1/3 of the overall code. [See the expanded code](https://github.com/rusterlium/rustler/files/3646858/expand.txt)

This could be reduced by generating the atoms only once by generating a helper module and referencing that module from encoders and decoders. I would be up to give that a try, but #234 should be merged before this in order to avoid merge issues.

PR #251, opened 2019-10-16
Refactor codegen

evnu:
The proc-macro code generation has a lot of duplicate code over multiple modules. This merge request refactors the modules by pulling common code into the `rustler_codegen::context::Context` struct and its functions. This is in part as a preparation for #241 in order to simplify de-duplicating the `field_atoms`.

### DIFF
--- a/rustler_codegen/src/context.rs
+++ b/rustler_codegen/src/context.rs
@@ -1,0 +1,138 @@
+use syn::{Lit, Meta, NestedMeta};
+
+use super::RustlerAttr;
+
+///
+/// A parsing context struct.
+///
+/// `Context` holds information usable for different codegen modules.
+///
+#[derive(Debug)]
+pub(crate) struct Context<'a> {
+    pub(crate) attrs: Vec<RustlerAttr>,
+    pub(crate) ident: &'a proc_macro2::Ident,
+    pub(crate) ident_with_lifetime: proc_macro2::TokenStream,
+}
+
+impl<'a> Context<'a> {
+    pub(crate) fn from_ast(ast: &'a syn::DeriveInput) -> Self {
+        let mut attrs: Vec<_> = ast
+            .attrs
+            .iter()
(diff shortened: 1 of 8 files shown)


evnu:
Rebased on master. @filmor do you want to give this PR a glance as well?

filmor:
I'll do that tonight, from a quick glance I only think you could drop the repeated `pub(crate)` in favour of `pub` in `Context` to make that one a bit less verbose (enough as the `struct` is already `pub(crate)`.

rustler_codegen/src/context.rs
```diff
@@ -0,0 +1,170 @@
...
+/// A parsing context struct.
+///
+/// `Context` holds information usable for different codegen modules.
+///
+pub(crate) struct Context<'a> {
+    pub(crate) attrs: Vec<RustlerAttr>,
+    pub(crate) ident: &'a proc_macro2::Ident,
+    pub(crate) ident_with_lifetime: proc_macro2::TokenStream,
+    pub(crate) variants: Option<Vec<&'a Variant>>,
+    pub(crate) struct_fields: Option<Vec<&'a Field>>,
```
filmor: As commented already, here it's sufficient to use `pub`.

rustler_codegen/src/context.rs
```diff
@@ -0,0 +1,170 @@
...
+}
+
+impl<'a> Context<'a> {
+    pub(crate) fn from_ast(ast: &'a syn::DeriveInput) -> Self {
```
filmor: Here as well, as the struct is `pub(crate)`.

filmor:
The rest looks good to me.

rustler_codegen/src/context.rs
evnu: 5b74eff

rustler_codegen/src/context.rs
evnu: 5b74eff

PR #252, opened 2019-10-16
Added Error::Term for returning `{:error, <term>}`

elbow-jason:
Hi! I added a variant to return any encoder as an error. I think could help ergonomics and help return more meaningful errors from our native Rustler interfaces to Elixir.

Feedback is appreciated. :)

### DIFF
--- a/rustler_tests/native/rustler_test/src/test_error.rs
+++ b/rustler_tests/native/rustler_test/src/test_error.rs
@@ -1,0 +1,45 @@
+use rustler::{Error, NifResult};
+
+mod atoms {
+    rustler::atoms! {
+        error,
+        should_be_a_raised_term_as_atom,
+        return_term_with_atom,
+        should_be_an_atom_wrapped_in_an_error_tuple,
+    }
+}
+
+#[rustler::nif]
+pub fn bad_arg_error() -> NifResult<()> {
+    Err(Error::BadArg)
+}
+
+#[rustler::nif]
+pub fn atom_str_error() -> NifResult<()> {
+    Err(Error::Atom("should_be_a_returned_atom"))
+}
+
+#[rustler::nif]
+pub fn raise_atom_error() -> NifResult<()> {
+    Err(Error::RaiseAtom("should_be_a_raised_atom"))
+}
+
+#[rustler::nif]
+pub fn raise_term_with_string_error() -> NifResult<()> {
+    Err(Error::RaiseTerm(Box::new(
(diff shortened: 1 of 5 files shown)


filmor:
Where would you return something arbitrary (i.e. not `{error, _}`) to indicate an error?

elbow-jason:
That's a great question. Idiomatically, I would only ever return `{:error, _}` as an error. 

Perhaps this function would be better as something like `ReturnErrorTuple(Box<Encoder>)` and return `{:error, _}`.

It would be pretty strange to return:
```rust 
ReturnTerm(Box::new((atoms::ok(), "some ok result wrapped in an error")))
```

@filmor is your concern something like the Rust code above?

rustler_tests/native/rustler_test/src/test_error.rs
```diff
@@ -0,0 +1,40 @@
+use rustler::{Env, Error, NifResult, Term};
+
+mod atoms {
+    rustler::rustler_atoms! {
```
scrogson: Please use the new `rustler::atoms!` macro here :wink:

filmor:
In line with the existing options, I think you should just add a `Error::Term(X)` that results in a returned error `{error, X}`.

elbow-jason:
@filmor Agreed; `{:error, term}` is preferrable.  I _should_ have time to get this done on Friday. Thanks for the feedback!

scrogson:
@elbow-jason...do you have time to finish this? If not, let me know and I can take it to the finish-line.

filmor:
Hmm, I just noticed that `Error::Atom("some_str")` actually returns `some_str` and this made me wonder whether it wouldn't be good to make this one also stricter by making it return `{error, some_str}`. The only halfway valid case to return a bare atom is forwarding of a `badmatch` or `badarg` and a totally invalid case is returning `Error::Atom("ok")`. Shouldn't be part of this PR, though.

elbow-jason:
@filmor `Error::Term` covers the use case you described for `Error::Atom`. I _think_ what you are recommending is the deprecation of `Error::Atom`.

filmor:
@elbow-jason No, I'm recommending a change of semantics for `Error::Atom` as I don't like it when libraries return a bare atom instead of `{error, Description}` as an error, but like @scrogson said, it's quite useful to have the possibility to return something like this without boxing :)

scrogson:
@elbow-jason now that #264 has been merged, feel free to fix up this PR so we can merge :+1:

Issue #160, opened 2018-06-21
Consider using the official Elixir autoformatter


walkr:
**Suggestion:**

I think it would be nice to use the Elixir formatter as it will make contributions easier and standardized with respect to code layout.

scrogson:
I'm personally not a big fan of the Elixir code formatter in its current state. It produces less readable code (IMHO). So I'm 👎 

I will leave it up to @hansihe

walkr:
Fair enough. 

Have you expressed your gripes regarding the formatter to the Elixir team? Maybe we can make it better.

I know typically a formatter cannot satisfy everyone's taste, yet I believe it provides formatting consistency across codebases. In addition, one could ensure better code readability, by properly breaking down the code into small, readable, functional blocks.

davydog187:
I added the formatter in #218 

I think enough time is passed where it has become a community standard, and it would be great for Rustler to adopt is as well.

scrogson:
Fixed in #253

PR #253, opened 2019-10-16
GitHub Actions

### LINKED ISSUE #121: Fix CI for real
The recently merged CI fixes feel a bit hacky. It would be nice to use `rust` as the language and test against all rust channels.

We need a better way to install `erlang` and `elixir`. Is `kerl` and `kiex` available inside all Travis builds?

scrogson:
An attempt to improve CI. Closes #121 and #160


rustler_mix/mix.exs
```diff
@@ -2,21 +2,25 @@ defmodule Rustler.Mixfile do
...
-     deps: deps(),
-     docs: [
-       extras: ["guides/Basics.md"],
-       source_url_pattern: "https://github.com/rusterlium/rustler/blob/master/rustler_mix/%{path}#L%{line}"
-     ],
-     package: package(),
-     description: description()]
+    [
+      app: :rustler,
+      # rustler_mix version
```
scrogson: This might cause issues with the release scripts, eh?

evnu:
@scrogson Can we also test older versions of Elixir? In Travis, we currently test against v1.6-v1.8.1.

scrogson:
> Can we also test older versions of Elixir? In Travis, we currently test against v1.6-v1.8.1.

Yeah. I was having weird issues with installing various versions of Erlang before. I’ll give it another shot.

scrogson:
Now testing Elixir v1.6...v1.9 on OTP 22 and testing against rust stable, beta, and nightly.

filmor:
More OTP versions would also be good (especially since the NIF API depends on that), 20 and 21 if possible.

filmor:
And is it really necessary to test more than `stable` rust? `beta` maybe, but I don't think nightly is required.

evnu:
This PR will also fix #160.

rustler_mix/mix.exs
evnu: Were fixes necessary to resolve this?

scrogson:
> More OTP versions would also be good (especially since the NIF API depends on that), 20 and 21 if possible.

Yeah, I've been having trouble when adding more OTP versions...

> And is it really necessary to test more than `stable` rust? `beta` maybe, but I don't think nightly is required.

This would be a question for @hansihe. I remember when I made the changes to Travis to fix our CI pipeline, it removed testing on nightly and Hans had voiced that he would like to test against nightly for something, but I don't recall the reason.

I do see your point though. This project targets stable, so that is probably all that matters.

scrogson:
Ok..I've just added OTP 21 and everything passed. Let me add 20 and see how it goes...

scrogson:
There seems to be an issue with installing Elixir with ESLs OTP 20.x due to missing packages as seem in this run: https://github.com/rusterlium/rustler/runs/266729233

I'm not sure how to resolve that with the current `actions/setup-elixir` action...

rustler_mix/mix.exs
scrogson: I learned that the `release.sh` script hasn't been working for a while. Probably after switching to workflows? We can revisit that in a separate PR.

scrogson:
Ok...so...looking at the `.travis.yml` on `master`...we don't actually test on OTP 20.x. In my last commit I added the file back and only added OTP 20 to the matrix. It fails because of `map_from_arrays`:

```
1) test map from arrays with non-matching length raises ArgumentError (RustlerTest.MapTest)
     test/map_test.exs:21
     Expected exception ArgumentError but nothing was raised
     code: assert_raise(ArgumentError, fn -> RustlerTest.map_from_arrays(keys, values) end)
     stacktrace:
       test/map_test.exs:24: (test)
```

So...we may need to either fix that, or just ignore it :smile:

scrogson:
Ok...all fixed. I don't necessarily like using both GH Actions and Travis, but this works. Thoughts?

.github/workflows/main.yml
```diff
@@ -0,0 +1,98 @@
...
+        working-directory: rustler_mix
+        run: mix format --check-formatted
+
+      - name: Check mix format (rustler_tests)
+        working-directory: rustler_tests
+        run: mix format --check-formatted
+
+  test:
+    name: OTP ${{matrix.otp}} / Elixir ${{matrix.elixir}} / Rust ${{matrix.rust}}
+    runs-on: ubuntu-latest
```
evnu: I wonder, is it possible to run this on macos and windows as well?

evnu: Ah, found my answer [here](https://github.com/actions/setup-elixir/blob/master/src/installer.js#L34):

> `'@actions/setup-elixir only supports Ubuntu Linux at this time'`

evnu:
> There seems to be an issue with installing Elixir with ESLs OTP 20.x due to missing packages as seem in this run.

https://github.com/actions/setup-elixir uses OTP 20 in their [matrix example](https://github.com/actions/setup-elixir#matrix-example). Maybe this is issue-worthy. Setting up OTP looks reasonably simple in [their repository](https://github.com/actions/setup-elixir/blob/master/src/install-otp-ubuntu). We should be able to replicate this problem.

> Thoughts?

I think we should stick with only one CI.  Github actions looks nice, and I believe that it will only grow over time, so I would tend to favor that. But, as long as we support OTP 20, this should also be tested on CI. Maybe we can substitute `actions/setup-elixir` with simple `apt` calls if we know how to fix the version problem?

filmor:
Drop OTP 20 then, 21 was a quite significant release anyhow (addition of `logger`). It should suffice to do manual testing on 20 before a release if we want explicit compatibility.

scrogson:
> Drop OTP 20 then, 21 was a quite significant release anyhow (addition of `logger`). It should suffice to do manual testing on 20 before a release if we want explicit compatibility.

Given the bug I found in the `map_from_arrays` by adding OTP 20...I think it's fine to keep the Travis check for now until we can find a fix for GitHub actions?

scrogson:
Let's get this merged so we can rebase all other PRs to ensure we get things formatted to reduce conflicts.

evnu:
W.r.t. the dpkg error when attempting to use [OTP 20](https://github.com/rusterlium/rustler/runs/266729233): Elixir's [compatibility table](https://hexdocs.pm/elixir/compatibility-and-deprecations.html#compatibility-between-elixir-and-erlang-otp) says that Elixir v1.9 should be able to work with OTP 20, but the ESL package requires OTP >= 21.3. This looks like a packaging error.

evnu:
> This looks like a packaging error.

It actually isn't a packaging error. Forcing installation of Elixir with `dpkg --force-all -i $package` won't work either:

```
...
"beam/beam_load.c(1863): Error loading module 'Elixir.IEx.CLI':\n  This BEAM file was compiled for a later version of the run-time system than 20.\n  To fix this, please recompile this module with an 20 compiler.\n  (Use of opcode 162; this emulator supports only up to 159.)\n"
2019-10-19 20:19:35 crash_report
    initial_call: {supervisor_bridge,user_sup,['Argument__1']}
...
```

So, while Elixir v1.9 is compatible to OTP 20, the .beam files actually aren't. The solution here should be to use Elixir v1.8.2 with OTP 20 in the matrix. Then, the `setup-elixir` action should work. @scrogson do you know if the build matrix can be specified to obey that requirement?

scrogson:
@evnu let me play around with the matrix a bit and see what control we have.

PR #256, opened 2019-10-18
Fix crash when compiling port

cristianberneanu:
A data section can contain multiple sets of values, instead of a single one.

### DIFF
--- a/rustler_mix/lib/mix/tasks/compile.rustler.ex
+++ b/rustler_mix/lib/mix/tasks/compile.rustler.ex
@@ -143,7 +143,11 @@ defmodule Mix.Tasks.Compile.Rustler do
   defp make_build_mode_flag(args, :debug), do: args ++ []
 
   defp get_name(cargo_data, section) do
-    get_in(cargo_data, [section, "name"])
+    case cargo_data[section] do
+      nil -> nil
+      values when is_map(values) -> values["name"]
+      values when is_list(values) -> Enum.find_value(values, & &1["name"])
+    end
   end
 
   def make_file_names(base_name, :lib) do


evnu:
@cristianberneanu I am running Elixir v1.9 here, but cannot reproduce a crash in `compile.rustler.ex`. Does `rustler_test` crash for you?

scrogson:
I think this means that your Cargo.toml is missing the `name` in the `[lib]` section?

cristianberneanu:
> @cristianberneanu I am running Elixir v1.9 here, but cannot reproduce a crash in compile.rustler.ex. Does rustler_test crash for you?

Sorry, I didn't test this properly.
I found this issue in my project, after upgrading to Elixir 1.9. But I also updated `rustler` to the latest version afterwards, to get rid of some warnings, and I assumed the problem was because of the new Elixir version.

The bug was in `rustler` and it appeared after this change: a96e2d442048fd8c76091f4b7d90155847a5e55d
Each section can have one or multiple set of values in it. I am compiling a Rust port driver in my project, which outputs a binary instead of a library, and I assume this is the reason it only happens for me.

I now made the proper fix and updated the PR title and description accordingly.
Thanks!

scrogson:
Ah I see. I believe this is because binaries are specified as `[[bin]]` right?

Also just for clarification, I believe you mean a port and not port driver? A port driver is more similar to a NIF (has a special C API) where as a port is when you communicate with an external program over stdin/stdout using the `Port.command` et. al.

cristianberneanu:
> Ah I see. I believe this is because binaries are specified as [[bin]] right?

I think so too.

> Also just for clarification, I believe you mean a port and not port driver? A port driver is more similar to a NIF (has a special C API) where as a port is when you communicate with an external program over stdin/stdout using the Port.command et. al.

Correct. I got confused in the terminology referring to ports, port drivers and NIF drivers.

scrogson:
@cristianberneanu it would be nice to update the `rustler_tests` project to add a hello world binary so we can test that this change works and that we don't break it again in the future. Is that something you can handle?

scrogson:
@cristianberneanu I've gone ahead and added tests and pulled in your change from this PR into my new one over at #261 

Are you cool with that?

scrogson:
#261 was merged. Thank you for your help uncovering that bug!

cristianberneanu:
> Are you cool with that? 

Yes, thanks for taking care of it.

PR #261, opened 2019-10-19
Fix Binary Compilation

scrogson:
This PR picks up from where #256 left off and adds some tests ~and tweaks the workspace configuration~.

~@evnu do the changes to the workspaces make sense?~


Cargo.toml
```diff
@@ -3,6 +3,4 @@ members = [
   "rustler",
   "rustler_codegen",
   "rustler_sys",
-  "rustler_tests/native/rustler_test",
```
evnu: I think I would not drop this. With this, running `cargo check` or similar in the root directory will also check the test crates, without having to run `mix test` in `rustler_tests` or `cargo check` for each of the test crates. For example, if something in `rustler_codegen` changes, problems are likely to be detected when running `cargo check` for `rustler_test`.

rustler_tests/native/deprecated_macros/Cargo.toml
```diff
@@ -11,3 +11,5 @@ crate-type = ["cdylib"]
 
 [dependencies]
 rustler = { path = "../../../rustler" }
+
+[workspace]
```
evnu: This is only a crate, but not a workspace. I think this is not needed.

rustler_tests/native/rustler_test/Cargo.toml
```diff
@@ -12,3 +12,5 @@ crate-type = ["cdylib"]
 [dependencies]
 lazy_static = "1.4"
 rustler = { path = "../../../rustler" }
+
+[workspace]
```
evnu: Same comment as [here](https://github.com/rusterlium/rustler/pull/261/files#r336749182), this is probably not needed?

evnu:
@scrogson I think you forgot to add the new crate? :)

rustler_tests/native/deprecated_macros/Cargo.toml
scrogson: I added this to stop it from failing to think it was in a workspace. Based on your input above, I will remove this and add to the root workspace. :+1:

Cargo.toml
scrogson: Makes sense :+1:

scrogson:
> I think you forgot to add the new crate? :)

I messed up the `.gitignore`

evnu:
@scrogson were you able to reproduce the error from #256 with the new crate? I still don't understand completely how the error is triggered. We use `[[bin]]` as well in our crates, but did not see any such error yet.

scrogson:
> were you able to reproduce the error from #256 with the new crate?

Yes. If you remove the first commit and try running this with the changes I've made, you should see the error.



> We use `[[bin]]` as well in our crates, but did not see any such error yet.

It looks like if there is a `[lib]` section in the `Cargo.toml` then mix task skips the `[[bin]]` section entirely (might be nice to support both compiling libs and bins in the same crate). I noticed that when I added the `[[bin]]` section to the `rustler_test` crate, there was no failure. But, it also didn't move the binary into the `priv/native` dir.

evnu:
This looks good then!

PR #262, opened 2019-10-19
Adjust the test matrix

scrogson:


scrogson:
@evnu this seems to do the trick! :tada:

.github/workflows/main.yml
```diff
@@ -51,18 +51,36 @@ jobs:
...
+          - erlang: 21.x
+            elixir: 1.9
+          - erlang: 21.x
+            elixir: 1.8
+          - erlang: 21.x
+            elixir: 1.7
+          - erlang: 21.x
+            elixir: 1.6
+
+          - erlang: 20.3.8.23
```
scrogson: I attempted to use `20.x` but that ends up failing to download `20.3.8` due to a 404 on Erlang Solutions.

evnu: > but that ends up failing to download 20.3.8 due to a 404 on Erlang Solutions.

According to [this page](https://www.erlang-solutions.com/resources/download.html) on ESL, 20.3.8 is indeed not made available. The `setup-elixir` action marks [20.3.8](https://github.com/actions/setup-elixir/blob/83d6ee76a6cdb237bfe8207ff77f091828cc5bcd/src/erlang-versions.txt#L41) as the most recent version for OTP 20, which is then picked up in [setup-elixir.js](https://github.com/actions/setup-elixir/blob/83d6ee76a6cdb237bfe8207ff77f091828cc5bcd/src/setup-elixir.js#L53). I don't know if semver is capable of picking up 20.3.8.23 as a valid version matching `20.X`. For now, I think it is OK to just use 20.3.8.23 here.

PR #264, opened 2019-10-23
NIF Codegen rework

scrogson:
I've got the new macros all working! :tada:

I'm pretty excited about this, but I'm pretty sure there are some opportunities for some clean up.

~One thing I've noticed is that if you specify any term which is has `decode` called on it, that function needs to return a `Result<T, E>`. It would be nice to avoid this if possible.~

Let me know what you think.

### DIFF
--- a/rustler/src/env.rs
+++ b/rustler/src/env.rs
@@ -23,7 +23,7 @@ pub struct Env<'a> {
 
 /// Two environments are equal if they're the same `NIF_ENV` value.
 ///
-/// A `Env<'a>` is equal to a `Env<'b>` iff `'a` and `'b` are the same lifetime.
+/// A `Env<'a>` is equal to a `Env<'b>` if and only if `'a` and `'b` are the same lifetime.
 impl<'a, 'b> PartialEq<Env<'b>> for Env<'a> {
     fn eq(&self, other: &Env<'b>) -> bool {
         self.env == other.env
--- a/rustler/src/export.rs
+++ b/rustler/src/export.rs
@@ -15,23 +15,16 @@
 /// The third argument is an `Option<fn(env: &Env, load_info: Term) -> bool>`. If this is
 /// `Some`, the function will execute when the NIF is first loaded by the BEAM.
 #[macro_export]
-#[deprecated(since = "0.22.0", note = "Please use `init!` instead.")]
+#[deprecated(since = "0.22.0", note = "Please use `rustler::init!` instead.")]
 macro_rules! rustler_export_nifs {
-    ($name:expr, [$( $exported_nif:tt ),+,], $on_load:expr) => {
-        $crate::init!($name, [$( $exported_nif ),*], $on_load);
-    };
-}
-
-#[macro_export]
-macro_rules! init {
     // Strip trailing comma.
     ($name:expr, [$( $exported_nif:tt ),+,], $on_load:expr) => {
-        $crate::init!($name, [$( $exported_nif ),*], $on_load);
+        $crate::rustler_export_nifs!($name, [$( $exported_nif ),*], $on_load);
     };
     ($name:expr, [$( $exported_nif:tt ),*], $on_load:expr) => {
         static mut NIF_ENTRY: Option<$crate::codegen_runtime::DEF_NIF_ENTRY> = None;
 
-        $crate::init!(internal_platform_init, ({
+        $crate::rustler_export_nifs!(internal_platform_init, ({
             // TODO: If an unwrap ever happens, we will unwind right into C! Fix this!
 
             extern "C" fn nif_load(
@@ -45,7 +38,7 @@ macro_rules! init {
             }
 
             const FUN_ENTRIES: &'static [$crate::codegen_runtime::DEF_NIF_FUNC] = &[
-                $($crate::init!(internal_item_init, $exported_nif)),*
+                $($crate::rustler_export_nifs!(internal_item_init, $exported_nif)),*
             ];
 
             let entry = $crate::codegen_runtime::DEF_NIF_ENTRY {
@@ -69,7 +62,7 @@ macro_rules! init {
     };
 
     (internal_item_init, ($nif_name:expr, $nif_arity:expr, $nif_fun:path)) => {
-        $crate::init!(internal_item_init, ($nif_name, $nif_arity, $nif_fun, $crate::schedule::SchedulerFlags::Normal))
+        $crate::rustler_export_nifs!(internal_item_init, ($nif_name, $nif_arity, $nif_fun, $crate::schedule::SchedulerFlags::Normal))
     };
     (internal_item_init, ($nif_name:expr, $nif_arity:expr, $nif_fun:path, $nif_flag:expr)) => {
         $crate::codegen_runtime::DEF_NIF_FUNC {
@@ -82,12 +75,9 @@ macro_rules! init {
                     argv: *const $crate::codegen_runtime::NIF_TERM)
                     -> $crate::codegen_runtime::NIF_TERM {
                         unsafe {
-                            $crate::init!(
+                            $crate::rustler_export_nifs!(
                                 internal_handle_nif_call, ($nif_fun, $nif_arity, env, argc, argv))
                         }
-                    //unsafe {
-                    //    $crate::codegen_runtime::handle_nif_call($nif_fun, $nif_arity, env, argc, argv)
-                    //}
                 }
                 nif_func
             },
--- a/rustler/src/lib.rs
+++ b/rustler/src/lib.rs
@@ -25,7 +25,8 @@
 #[macro_use(enif_snprintf)]
 extern crate rustler_sys;
 
-mod wrapper;
+#[doc(hidden)]
+pub mod wrapper;
 
 #[doc(hidden)]
 pub mod codegen_runtime;
@@ -62,7 +63,13 @@ pub use crate::error::Error;
 pub mod r#return;
 pub use crate::r#return::Return;
 
+#[doc(hidden)]
+mod nif;
+pub use nif::Nif;
+
 pub type NifResult<T> = Result<T, Error>;
 
 #[cfg(feature = "derive")]
-pub use rustler_codegen::{NifMap, NifRecord, NifStruct, NifTuple, NifUnitEnum, NifUntaggedEnum};
+pub use rustler_codegen::{
+    init, nif, NifMap, NifRecord, NifStruct, NifTuple, NifUnitEnum, NifUntaggedEnum,
+};
--- a/rustler_codegen/src/init.rs
+++ b/rustler_codegen/src/init.rs
@@ -7,38 +7,57 @@ use syn::{Expr, Ident, Result, Token};
 
 #[derive(Debug)]
 pub struct InitMacroInput {
-    module: syn::Lit,
+    name: syn::Lit,
     funcs: syn::ExprArray,
-    on_load: Option<TokenStream>,
+    load: TokenStream,
 }
 
 impl Parse for InitMacroInput {
     fn parse(input: ParseStream) -> Result<Self> {
-        let module = syn::Lit::parse(input)?;
+        let name = syn::Lit::parse(input)?;
         let _comma = <syn::Token![,]>::parse(input)?;
         let funcs = syn::ExprArray::parse(input)?;
-        let on_load = if input.peek(Token![,]) {
-            let _comma = <Token![,]>::parse(input)?;
-            Some(TokenStream::parse(input)?)
-        } else {
-            let none = Ident::new("None", Span::call_site());
-            Some(quote!(#none))
-        };
+        let options = parse_expr_assigns(input);
+        let load = extract_option(options, "load");
 
-        Ok(InitMacroInput {
-            module,
-            funcs,
-            on_load,
-        })
+        Ok(InitMacroInput { name, funcs, load })
     }
+}
+
+fn parse_expr_assigns(input: ParseStream) -> Vec<syn::ExprAssign> {
+    let mut vec = Vec::new();
+
+    while let Ok(_) = <Token![,]>::parse(input) {
+        match syn::ExprAssign::parse(input) {
+            Ok(expr) => vec.push(expr),
+            Err(err) => panic!("{} (i.e. `load = load`)", err),
+        }
+    }
+    vec
+}
+
+fn extract_option(args: Vec<syn::ExprAssign>, name: &str) -> TokenStream {
+    for syn::ExprAssign { left, right, .. } in args.into_iter() {
+        if let syn::Expr::Path(syn::ExprPath { path, .. }) = &*left {
+            if let Some(ident) = path.get_ident() {
+                if *ident == name {
+                    let value = *right.clone();
+                    return quote!(Some(#value));
+                }
+            }
+        }
+    }
+
+    let none = Ident::new("None", Span::call_site());
+    quote!(#none)
 }
 
 impl Into<proc_macro2::TokenStream> for InitMacroInput {
     fn into(self) -> proc_macro2::TokenStream {
-        let name = self.module;
+        let name = self.name;
         let num_of_funcs = self.funcs.elems.len();
         let funcs = nif_funcs(self.funcs.elems);
-        let on_load = self.on_load;
+        let load = self.load;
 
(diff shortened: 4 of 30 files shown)


rustler_codegen/src/init.rs
```diff
@@ -0,0 +1,114 @@
...
+        let name = self.module;
+        let num_of_funcs = self.funcs.elems.len();
+        let funcs = nif_funcs(self.funcs.elems);
+        let on_load = self.on_load;
+
+        let inner = quote! {
+            static mut NIF_ENTRY: Option<rustler::codegen_runtime::DEF_NIF_ENTRY> = None;
+            const FUN_ENTRIES: &'static [rustler::codegen_runtime::DEF_NIF_FUNC] = &[
+                #funcs
+            ];
```
scrogson: I think we can get rid of this somehow?

rustler_tests/native/rustler_test/src/test_atom.rs
```diff
@@ -1,28 +1,27 @@
...
-    args[0].atom_to_string()
+#[rustler::nif]
+pub fn atom_to_string(atom: Term) -> NifResult<String> {
+    atom.atom_to_string()
 }
 
-pub fn atom_equals_ok<'a>(_env: Env<'a>, args: &[Term<'a>]) -> bool {
-    atoms::ok() == args[0]
+#[rustler::nif]
+pub fn atom_equals_ok(atom: Atom) -> NifResult<bool> {
```
scrogson: Something is weird here and I couldn't quite figure out why, but some of these require `NifResult<T>` while others don't...

rustler/src/env.rs
```diff
@@ -23,7 +23,7 @@ pub struct Env<'a> {
 
 /// Two environments are equal if they're the same `NIF_ENV` value.
 ///
-/// A `Env<'a>` is equal to a `Env<'b>` iff `'a` and `'b` are the same lifetime.
+/// A `Env<'a>` is equal to a `Env<'b>` if `'a` and `'b` are the same lifetime.
```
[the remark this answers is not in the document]
scrogson: Ah thanks for pointing that out. As this is not something that was intuitive to me, I've updated it to "if and only if" to avoid user confusion in the future.

rustler/src/export.rs
```diff
@@ -17,21 +17,14 @@
 #[macro_export]
 #[deprecated(since = "0.22.0", note = "Please use `init!` instead.")]
```
evnu: Maybe the deprecation warning should advise to use `rustler::init!` (full path to access the function) instead to make it easier to find.

rustler_tests/native/rustler_test/src/test_term.rs
```diff
@@ -17,11 +10,21 @@ mod atoms {
     }
 }
 
-pub fn term_eq<'a>(_env: Env<'a>, args: &[Term<'a>]) -> bool {
-    args[0] == args[1]
+#[rustler::nif]
+pub fn term_debug(term: Term) -> NifResult<String> {
```
evnu: On my machine, this doesn't need `NifResult`.

evnu: Or, the wrapping with `Ok(..)` is not necessary here if `NifResult` is dropped.

rustler_tests/native/rustler_test/src/test_atom.rs
evnu: I made an attempt to get rid of some of those `NifResult` types on a [test branch](https://github.com/evnu/rustler/tree/rustler_macros-wrap-ok).

In this example here (`atom_equals_ok`), the problem is that the expanded code first tries to decode an `Atom`, which can fail. Thus, the expanded code requires that the closure in `catch_unwind` returns a type `Result<_, rustler::error::Error>`, where the error type is defined by the decoder. Now, if the test function here returns only `bool`, we have a type mismatch. We want to return a `bool` from the closure, but we are required to return a result. We can work around this issue by always returning `Ok(nif result)` from the closure instead. This appears simple at first, but we must then ensure to remove this extraneous `Ok(_)` again when handling the result. If we do not do that, we cannot use `NifResult` anymore as a return type for a `rustler::nif` function.

@scrogson Does the explanation make sense? The diff in my test branch might be simpler to follow than this explanation :) On that particular branch, it is now possible to return bool directly from `atom_equals_ok`.

evnu:
> One thing I've noticed is that if you specify any term which is has decode called on it, that function needs to return a Result<T, E>.

I am not sure if I understand this correctly, but I think my [comment here](https://github.com/rusterlium/rustler/pull/264#discussion_r339347098) might cover this.

rustler_tests/native/rustler_test/src/test_atom.rs
evnu: Note that this was a bit tedious to debug. To figure out what goes wrong, I used [cargo-expand](https://github.com/dtolnay/cargo-expand) on the current version of the code and replaced the code for `atom_equals_ok` with the expanded version. Then, I hacked and slashed around in the expanded code until I was able to reproduce the error, and then played around with wrapping the result in `Ok`. Alas, `cargo-expand` only works if the code compiles, so there is some back-and-forth needed between fixing something in codegen and debugging it in the expanded code.

rustler_codegen/src/init.rs
```diff
@@ -0,0 +1,110 @@
...
+                num_of_funcs: #num_of_funcs as rustler::codegen_runtime::c_int,
+                funcs: [#funcs].as_ptr(),
+                load: {
+                    extern "C" fn nif_load(
+                        env: rustler::codegen_runtime::NIF_ENV,
+                        _priv_data: *mut *mut rustler::codegen_runtime::c_void,
+                        load_info: rustler::codegen_runtime::NIF_TERM
+                    ) -> rustler::codegen_runtime::c_int {
+                        unsafe {
+                            rustler::codegen_runtime::handle_nif_init_call(#on_load, env, load_info)
```
evnu: I think the TODO from [here](https://github.com/rusterlium/rustler/pull/264/files#diff-60e84eda516f7ecf0da92f510a594394R28) might still apply to this? If so, we should copy the TODO, or maybe open an issue to fix it.

evnu: The TODO:

```
// TODO: If an unwrap ever happens, we will unwind right into C! Fix this!
```

rustler_codegen/src/nif.rs
```diff
@@ -0,0 +1,205 @@
...
+    let arity = arity(inputs.clone());
+    let decoded_terms = extract_inputs(inputs.clone());
+    let argument_names = create_function_params(inputs.clone());
+    let erl_func_name = extract_attr_value(args.clone(), "name")
+        .map(|ref n| syn::Ident::new(n, Span::call_site()))
+        .unwrap_or_else(|| name.clone());
+
+    quote! {
+        #[allow(non_camel_case_types)]
+        pub struct #name;
```
evnu: This could also be `pub enum #name{}`. With that, it is impossible to create an instance of that type. As the type is "only" used to implement `rustler::Nif` on it, and the type is not passed around, this could make sense.

rustler_codegen/src/nif.rs
```diff
@@ -0,0 +1,205 @@
...
+use syn::punctuated::Punctuated;
+use syn::token::Comma;
+
+pub fn transcoder_decorator(args: syn::AttributeArgs, fun: syn::ItemFn) -> TokenStream {
+    let sig = &fun.sig;
+    let name = &sig.ident;
+    let inputs = &sig.inputs;
+    let flags = schedule_flag(args.to_owned());
+    let function = fun.to_owned().into_token_stream();
+    let arity = arity(inputs.clone());
```
evnu: `arity()` could also take `&Punctuated<_,_>`. Then, cloning is not needed. This is also possible with `extract_inputs` and `create_function_params`.

rustler_codegen/src/nif.rs
```diff
@@ -0,0 +1,205 @@
+use proc_macro2::{Span, TokenStream};
+use quote::{quote, ToTokens};
+use syn::punctuated::Punctuated;
+use syn::token::Comma;
+
+pub fn transcoder_decorator(args: syn::AttributeArgs, fun: syn::ItemFn) -> TokenStream {
+    let sig = &fun.sig;
+    let name = &sig.ident;
+    let inputs = &sig.inputs;
+    let flags = schedule_flag(args.to_owned());
```
evnu: `schedule_flag` could also use `&args` (if `extract_attr_value` did the same). This seems to work here, but maybe there is a better reason to not use references here or below?

rustler_tests/native/rustler_test/src/test_atom.rs
scrogson: Yeah, I had tried a few similar things but I think the missing link for me was specifying the actual return type in the `catch_unwind`. I finally just settled to post a PR and hopefully figure it out later. It was nearly 2am 😆

evnu:
@scrogson I really like this PR, great work! :)

rustler_codegen/src/nif.rs
scrogson: I had all this at first, but then clippy was complaining:

```
writing `&Vec<_>` instead of `&[_]` involves one more reference and cannot be used with non-Vec-based slices.
```

evnu: Interesting, clippy does not complain here when I try that (clippy 0.0.212 (3aea860 2019-09-03)). Feel free to resolve here and the related comments!

rustler_codegen/src/nif.rs
scrogson: This did not work.

scrogson:
> I really like this PR, great work! :)

@evnu, thank you. This has been on my mind for quite a long time. It feels great to finally get it working. Thanks for your review!

scrogson:
When looking into the failures with the `rustler_mix_test` I realized that the `init!` macro doesn't work without specifying the `on_load` function as either `None` or `Some(fun)`.

```
error: expected expression, found `,`
 --> src/lib.rs:6:1
  |
6 | rustler::init!("Elixir.RustlerMixTest", [add]);
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected expression
```

Running `cargo expand` on the code reveals this output for the `nif_load` body:

```rust
extern "C" fn nif_load(
    env: rustler::codegen_runtime::NIF_ENV,
    _priv_data: *mut *mut rustler::codegen_runtime::c_void,
    load_info: rustler::codegen_runtime::NIF_TERM,
) -> rustler::codegen_runtime::c_int {
    unsafe {
        (/*ERROR*/)
    }
}
```

@evnu do you have any ideas how to fix that?

filmor:
I'll have a look at the code tomorrow, but one more suggestion: Now that this is an actual `proc_macro`, I think we should support a syntax like this instead:

    rustler::init!("mod", entries, on_load = something)

This opens up the possibility of adding more parameters in the future (e.g. I'd really like to be able to allow a `reload`.

scrogson:
> This opens up the possibility of adding more parameters in the future (e.g. I'd really like to be able to allow a `reload`.

Yeah, I like that. I'll work on that tonight.

rustler_codegen/src/init.rs
```diff
@@ -0,0 +1,111 @@
...
+impl Into<proc_macro2::TokenStream> for InitMacroInput {
+    fn into(self) -> proc_macro2::TokenStream {
+        let name = self.module;
+        let num_of_funcs = self.funcs.elems.len();
+        let funcs = nif_funcs(self.funcs.elems);
+        let on_load = self.on_load;
+
+        let inner = quote! {
+            static mut NIF_ENTRY: Option<rustler::codegen_runtime::DEF_NIF_ENTRY> = None;
+            use rustler::Nif;
```
evnu: I think this `use` is not used below?

rustler_codegen/src/init.rs
```diff
@@ -0,0 +1,111 @@
...
+                funcs: [#funcs].as_ptr(),
+                load: {
+                    extern "C" fn nif_load(
+                        env: rustler::codegen_runtime::NIF_ENV,
+                        _priv_data: *mut *mut rustler::codegen_runtime::c_void,
+                        load_info: rustler::codegen_runtime::NIF_TERM
+                    ) -> rustler::codegen_runtime::c_int {
+                        unsafe {
+                            // TODO: If an unwrap ever happens, we will unwind right into C! Fix this!
+                            rustler::codegen_runtime::handle_nif_init_call(#on_load, env, load_info)
```
evnu: WRT https://github.com/rusterlium/rustler/pull/264#issuecomment-548420083:

When `self.on_load` is `None` (https://github.com/rusterlium/rustler/pull/264/files#diff-8daf9e0986d966ebc2c4878e8a772b5fR40), this will not be expanded. It seems that we must pass-in a Tokenstream here, even if we do not have an `on_load` function.

evnu: We could parse `on_load` as follows, which would make this work here:

```rust
        let on_load = if input.peek(Token![,]) {
            let _comma = <Token![,]>::parse(input)?;
            TokenStream::parse(input)?
        } else {
            let none_ident = Ident::new("None", Span::call_site());
            Some(quote!{#none_ident})
        };
```

We should then probably change the type of `on_load` from `Option<TokenStream>` to `TokenStream`, as it will be `Some(..)` anyways.

evnu:
> Running cargo expand on the code reveals this output for the nif_load body:

See https://github.com/rusterlium/rustler/pull/264/files#r341824704

We need to encode `None` as a TokenStream. Otherwise, it seems that expanding it results in the whole call to `handle_nif_init_call()` being dropped. I don't know why this happens (some property of `quote!{}` maybe?), but quoting `None` manually works.

rustler_codegen/src/init.rs
evnu: Ah, it is used to enable retrieving the associated constants.

evnu:
@scrogson see https://github.com/evnu/rustler/commit/1794b330d33c63a6f035cf56b2929d61d9b6f60b for an idea on how to fix the problem with `on_load`. See https://github.com/evnu/rustler/commit/5e22350b4211fbf226b26bb07e76c9d1cc223a0b for a proposal to fix the template.

rustler_tests/native/rustler_test/src/test_env.rs
```diff
@@ -65,5 +64,5 @@ pub fn sublists<'a>(env: Env<'a>, args: &[Term<'a>]) -> NifResult<Term<'a>> {
         });
     });
 
-    Ok(atom::ok().to_term(env))
+    Ok(atom::ok())
```
filmor: Maybe just return `Atom` here?

rustler/src/lib.rs
```diff
@@ -62,7 +63,13 @@ pub use crate::error::Error;
...
+#[doc(hidden)]
+mod nif;
+pub use nif::Nif;
+
 pub type NifResult<T> = Result<T, Error>;
 
 #[cfg(feature = "derive")]
-pub use rustler_codegen::{NifMap, NifRecord, NifStruct, NifTuple, NifUnitEnum, NifUntaggedEnum};
+pub use rustler_codegen::{
+    init, nif, NifMap, NifRecord, NifStruct, NifTuple, NifUnitEnum, NifUntaggedEnum,
```
filmor: Shouldn't `init!` and `nif!` be included unconditionally?

rustler_tests/native/rustler_test/src/test_env.rs
scrogson: This function uses `?` so it has to return a `Result`.

rustler/src/lib.rs
scrogson: Perhaps they should. But right now, we have `rustler_codegen` as an optional dependency behind the `derive` feature flag (which is default).

Maybe we don't need this as a feature at all? I think it was just put in place because previously `rustler_codegen` had to be pulled in manually by users and in the latest version, we pulled it in by default.

But now that `init` and `nif` are primary macros, I think we should probably remove it.

filmor: Yep, can be done in a separate PR, though.

filmor:
@evnu Can you have a last look? In my opinion this is good to go.

Tremendous work @scrogson, this makes the interface so much simpler :)

PR #267, opened 2019-10-27
Fix replacing rustler version in release script

evnu:
`mix format` in #253 broke changing the rustler version in `rustler_mix/mix.exs` done by `release.sh` ([here](https://github.com/rusterlium/rustler/pull/253/files#diff-107359545ec07acf37f512453c7469e0)).


evnu:
On a related note, we might want to start tagging releases. With that, we could use the tag for the `source_url_pattern` for the documentation:

```elixir
        source_url_pattern:
          "https://github.com/rusterlium/rustler/blob/A_VERSION_TAG/rustler_mix/%{path}#L%{line}"
```

That way, source links in the released Elixir documentation on hex.pm will remain useful even if the files change on master.

scrogson:
> On a related note, we might want to start tagging releases.

Yes! This is something that I want to start doing in the next release. Along with adding the changes from the changelog as release notes.

PR #266, opened 2019-10-27
Test rustler_mix generated files

evnu:
This pull request adds a shell script to test that the files generated by `rustler_mix` work when used to create a new crate. I fixed some small issues while implementing the script, and extended the mix task to allow specifying the module, name and otp app on the command line.

PR #269, opened 2019-10-27
rustler_mix: Fix template test on Elixir < 1.9

evnu:
Another fix for #266 (d7552b4). This also adds the event `pull_request` to github actions in order to try to test this PR before it hits master. We can drop this commit when CI succeeded.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -50,6 +50,27 @@ jobs:
         working-directory: rustler_tests
         run: mix format --check-formatted
 
+  rustler_mix_test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+
+      - name: Install Rust stable toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          override: true
+
+      - name: Install Erlang/Elixir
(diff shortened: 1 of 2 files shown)


evnu:
@scrogson Can you take a look why CI is failing? The error messages are not particularly helpful:

image

Maybe there are two many concurrent jobs and they are cancelled?

scrogson:
Maybe this? https://github.com/rusterlium/rustler/pull/269/checks?check_run_id=276817469#step:6:64

scrogson:
We can pull this check out into its own yml file and only run on stable and latest Erlang/Elixir. This is what I did for the Clippy and Format checks.

scrogson:
Sorry I was wrong about the separation file. But we can either pull it into its own section or create a separate file.

evnu:
I pulled the test into its own step with 208fdd0, which worked on CI. I think 6e8a205 already fixed it, but CI somehow stumbled over the large number of builds..(?) In any case, running this test only once should be sufficient to pick up simple errors where the templates are not updated.

evnu:
Alright, this seems to work now. The original fix in this PR was indeed not enough. @scrogson should I leave 158d4f6 in or drop it?

PR #272, opened 2019-11-04
Implement TryFrom<&[u8]> for OwnedBinary

scrogson:


filmor:
1. With this one there are now essentially 3 versions of the same code: `OwnedBinary::from_unowned`, `OwnedBInary::realloc_or_copy` and this one. Can you consolidate them to a single implementation of "filling `OwnedBinary`"?
2. Wouldn't an implementation of `TryFrom<Borrow<[u8]>>` make more sense to catch also cases like `Vec<u8>`?

evnu:
> OwnedBinary::from_unowned, OwnedBInary::realloc_or_copy and this one.

It would be nice if those functions also used `write_all()` like the function proposed here. Refactoring sounds reasonable.

> TryFrom<Borrow<[u8]>>

The Rust documentation on [Borrow](https://doc.rust-lang.org/std/borrow/trait.Borrow.html) advises to use `AsRef` if only a reference to a related type is needed, maybe this would be better?

scrogson:
I tried to do this, but unfortunately it appears that the compile doesn't like this:

```
error[E0119]: conflicting implementations of trait `std::convert::TryFrom<_>` for type `types::binary::OwnedBinary`:
  --> rustler/src/types/binary.rs:17:1
   |
17 | / impl<T> TryFrom<T> for OwnedBinary
18 | | where
19 | |     T: AsRef<[u8]>,
20 | | {
...  |
31 | |     }
32 | | }
   | |_^
   |
   = note: conflicting implementation in crate `core`:
           - impl<T, U> std::convert::TryFrom<U> for T
             where U: std::convert::Into<T>;

error: aborting due to previous error

For more information about this error, try `rustc --explain E0119`.
```
I believe that we'd need specialization for this to work?

PR #275, opened 2019-11-11
Refactor compilation

scrogson:
Currently the mix compiler adds additional work for users in order to get started. The changes proposed in this PR aims to make this go away completely and allow users to turn off compilation completely in favor of loading a pre-compiled binary.

**Pros**

* Users no longer need to add the `rustler` compiler in `mix.exs`
* Users no longer need to add the `rustler_crates` option in `mix.exs`
* Crates are automatically compiled when the associated NIF module is compiled.
* Recompilation only happens when one of Rust files or the NIF module are updated (see [`@external_resource`](https://hexdocs.pm/elixir/Module.html#module-external_resource)).

**Cons**

~* Currently, when changes are made to the rust files in the project, `mix compile` does not recompile unless the NIF module has been saved.~
* This change breaks existing binary compilation (but I think this can be fixed pretty easily).

There might be more cons than I can think of right now...

**TODO**

- [x] Revamp the docs
- [x] Support compiling rust binaries
- [x] Support loading pre-compiled libs/binaries

### DIFF
--- a/rustler_mix/README.md
+++ b/rustler_mix/README.md
@@ -14,7 +14,7 @@ This package is available on [`hex.pm`](https://hex.pm/packages/rustler). To ins
 
 ```elixir
 def deps do
-  [{:rustler, "~> 0.21.0"}]
+  [{:rustler, "~> 0.21.0", runtime: false}]
 end
 ```
 
@@ -54,7 +54,9 @@ loaded.
 
 ```elixir
 defmodule MyProject.MyModule do
-  use Rustler, otp_app: :my_app, crate: :my_crate
+  use Rustler,
+    otp_app: :my_app,
+    crate: :my_crate
 
   # When loading a NIF module, dummy clauses for all NIF function are required.
   # NIF dummies usually just error out when called when the NIF is not loaded, as that should never normally happen.
--- a/rustler_mix/README.md
+++ b/rustler_mix/README.md
@@ -3,8 +3,7 @@
 This is the Mix package for [rustler](https://github.com/rusterlium/rustler), a library to write Erlang NIFs in
 safe Rust code. Here, we provide the basic functionality to use Rustler from Elixir:
 
-* A task to generate a new crate to write NIFs (`mix help rustler.new`)
-* A task to compile NIFs written in Rust (`mix help compile.rustler`)
+- A task to generate a new crate to write NIFs (`mix help rustler.new`)
 
 See below for information on how to install this, which options are exposed through the configuration, and how to
 load a NIF.
@@ -22,10 +21,8 @@ end
 Then,
 
 1. Run `mix deps.get` to fetch the dependency.
-2. Run `mix rustler.new` and follow the instructions to generate the boilerplate for your NIF.
-3. Enable the `:rustler` mix compiler by adding `compilers: [:rustler] ++ Mix.compilers(),` to the `project` section of your `mix.exs`.
-4. Add a configuration entry to the `rustler_crates` section of your `mix.exs`. [See below](#crate-configuration).
-5. Load the NIF in your program. [See below](#loading-the-nif).
+1. Run `mix rustler.new` and follow the instructions to generate the boilerplate for your NIF.
+1. Load the NIF in your program. [See below](#loading-the-nif).
 
 ## Crate configuration
 
@@ -40,40 +37,12 @@ The NIF configuration may contain the following entries:
   - `{:bin, "path"}` - Use `path` as the cargo command. This is not portable, and you should not normally use this.
 - `default_features` (true default) - Boolean indicating if you want the NIF built with or without default cargo features.
 - `features` ([] default) - List of binaries indicating what cargo features you want enabled when building.
-- `mode` (:release default) - Indicates what cargo build flavor you want.
+- `target` (nil default) - Specify which build target to compile for. See .
+- `mode` - Indicates what cargo build flavor to compile with. The default
+  depends on the `Mix.env()`. When `:prod` the crate will be compiled in
+  `release` mode. Otherwise it will be compiled in `debug` mode.
   - `:release` - Optimized build, normally a LOT faster than debug.
   - `:debug` - Unoptimized debug build with debug assertions and more.
-
-When you are done, the project section might look something like this:
-
-```elixir
-def project do
-  [app: :my_app,
-   version: "0.1.0",
-   compilers: [:rustler] ++ Mix.compilers(),
-   rustler_crates: [my_crate: []],
-   deps: deps()]
-end
-```
-
-### Conditionally setting the mode
-
-The `rustc` mode defaults to `release`, but if you wish to compile your crate
-in `debug` mode for `dev`/`test` but want `release` mode for `prod`:
-
-```elixir
-def project do
-  [app: :my_app,
-   version: "0.1.0",
-   compilers: [:rustler] ++ Mix.compilers(),
-   rustler_crates: [
-     my_crate: [
-       mode: (if Mix.env() == :prod, do: :release, else: :debug)
-     ]
-   ],
-   deps: deps()]
-end
-```
 
 ## Loading the NIF
 
--- a/rustler_mix/README.md
+++ b/rustler_mix/README.md
@@ -26,23 +26,10 @@ Then,
 
 ## Crate configuration
 
-The `rustler_crates` configuration is a keyword list mapping the crate name (an atom) to the NIF configuration (another keyword list).
-The NIF configuration may contain the following entries:
+The Rust crate compilation can be controlled via Mix compile-time configuration in `config/config.exs`.
+See [configuration options](https://hexdocs.pm/rustler/Rustler.html#module-configuration-options) for more
+details.
 
-- `path` - The path to the crate directory relative to the project root
-  (default: `native/<crate-name>`)
-- `cargo` (:system default) - The rust/cargo version to build the NIF with. May be one of the following:
-  - `:system` - Use the version installed on the system.
-  - `{:rustup, "rust-version"}` - Use `rustup` to compile the NIF with a specific version.
-  - `{:bin, "path"}` - Use `path` as the cargo command. This is not portable, and you should not normally use this.
-- `default_features` (true default) - Boolean indicating if you want the NIF built with or without default cargo features.
-- `features` ([] default) - List of binaries indicating what cargo features you want enabled when building.
-- `target` (nil default) - Specify which build target to compile for. See .
-- `mode` - Indicates what cargo build flavor to compile with. The default
-  depends on the `Mix.env()`. When `:prod` the crate will be compiled in
-  `release` mode. Otherwise it will be compiled in `debug` mode.
-  - `:release` - Optimized build, normally a LOT faster than debug.
-  - `:debug` - Unoptimized debug build with debug assertions and more.
 
 ## Loading the NIF
 
--- a/rustler_mix/lib/rustler.ex
+++ b/rustler_mix/lib/rustler.ex
@@ -28,24 +28,40 @@ defmodule Rustler do
   Either of the above options can be passed directly into the `use` macro like so:
 
       defmodule MyNIF do
-        use Rustler, otp_app: :my_nif, crate: :some_other_crate, load_data: :something
+        use Rustler,
+          otp_app: :my_nif,
+          crate: :some_other_crate,
+          load_data: :something
       end
 
   """
 
   defmacro __using__(opts) do
     quote bind_quoted: [opts: opts] do
-      @on_load :__init__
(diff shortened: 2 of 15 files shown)


rustler_mix/lib/rustler/compiler.ex
```diff
@@ -1,58 +1,42 @@
...
       )
 
+    priv_dir =
+      otp_app
+      |> :code.priv_dir()
+      |> to_string()
+
+    File.mkdir_p!(priv_dir())
+
+    lib_path = String.to_charlist("#{priv_dir}/native/lib#{crate}")
```
filmor: Should we decide on a common scheme here for `rebar3_cargo` and this? What I'm doing there (right now) is `#{priv_dir}/crates/#{name}/#{version}/#{mode}/` and drop all build results in there.

hansihe:
I like this in general.

The only thing I want to be careful with is binary precompilation. We don't want to put ourselves in a position where libraries on hex using rustler don't work on some platforms because they only include binaries for some platforms. As long as we keep compilation as a fallback, this should be fine.

rustler_mix/priv/templates/basic/README.md
```diff
@@ -2,8 +2,6 @@
 
 ## To build the NIF module:
 
-- Make sure your projects `mix.exs` has the `:rustler` compiler listed in the `project` function: `compilers: [:rustler] ++ Mix.compilers()` If there already is a `:compilers` list, you should append `:rustler` to it.
-- Add your crate to the `rustler_crates` attribute in the `project function. [See here](https://hexdocs.pm/rustler/basics.html#crate-configuration).
 - Your NIF will now build along with your project.
```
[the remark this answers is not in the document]
scrogson: Yep, docs are still a TODO item.

rustler_mix/lib/rustler/compiler.ex
scrogson: So...the way this works currently is...`cargo` compiles everything in `_build/#{Mix.env()}/lib/#{otp_app}/native/#{crate}`.

Then only the necessary artifacts are moved into `priv/native`.

scrogson:
> We don't want to put ourselves in a position where libraries on hex using rustler don't work on some platforms because they only include binaries for some platforms.

Yep, I'm in agreement with this.

When I say pre-compiled, I'm talking about something that the user would download or build themselves. Then place that pre-compiled artifact in the `priv/native` directory of their application, and use `config.exs` to configure the NIF module to skip compilation and load the file from the specified location.

```elixir
# config.exs
config :some_nif, SomeNIF,
  skip_compilation: true,
  load_from: {:my_app, "libsome_nif.so"}
```

rustler_mix/lib/rustler/compiler.ex
filmor: That will just effectively prevent you from being able to reload, as some OSes won't load a changed library from the same location. Windows will even prevent overwriting the in-use DLL. And to be able to reliably get the build target and version you will need to parse `cargo`'s output (not a huge deal, though, it's very simple, see https://github.com/filmor/rebar3_cargo/blob/cde49c2a783e523090d88a77549228efd81d110f/src/rebar3_cargo_compile_prv.erl#L82-L102).

scrogson:
Ok...I've made some more progress on the functional aspects of this PR.

I've made it so that this will also move compiled binaries included in the crate, but...this PR does not support standalone binary crates anymore.

I think we can still support this by adding some sort of `binary: true` option to the `use Rustler` options. But this does require the user to create a module.

Given that Rustler's use-case is focused on NIFs, I don't know how much effort we should go through to maintain it. Thoughts?

rustler_mix/README.md
```diff
@@ -40,41 +37,13 @@ The NIF configuration may contain the following entries:
   - `{:bin, "path"}` - Use `path` as the cargo command. This is not portable, and you should not normally use this.
 - `default_features` (true default) - Boolean indicating if you want the NIF built with or without default cargo features.
 - `features` ([] default) - List of binaries indicating what cargo features you want enabled when building.
-- `mode` (:release default) - Indicates what cargo build flavor you want.
+- `target` (nil default) - Specify which build target to compile for. See .
```
evnu: > See .

I think there is a link missing here?

rustler_mix/lib/rustler.ex
```diff
@@ -18,34 +18,98 @@ defmodule Rustler do
...
     * `:crate` - the name of the Rust crate (as an atom), if different from your
       `otp_app` value. If you have more than one crate in your project, you will
       need to be explicit about which crate you intend to use.
 
+    * `:default_features` - a boolean to specify whether the crate's default features
+      should be used.
+
+    * `:env` - Specify a list of environment variables when envoking the compiler.
+
+    * `:feature` - a list of features to enable when compiling the crate.
```
evnu: Is this a comma-separated list?

evnu: Ah, ignore this, I was thinking in the context of the command line, not the config. :)

scrogson: Ah....I do think this should be `features` (plural) though.

rustler_mix/README.md
scrogson: Yep...thanks.

evnu:
@scrogson Is this obsolete now? https://github.com/rusterlium/rustler/blob/master/rustler_mix/lib/rustler/compiler/messages.ex#L27

evnu:
@scrogson When a user forgets to remove `:rustler` from the list of compilers but the `rustler_crates` option has been removed, the following error is thrown:

```
** (Mix) Missing required :rustler_crates option in mix.exs.
```

rustler_mix/lib/rustler.ex
```diff
@@ -16,36 +16,100 @@ defmodule Rustler do
...
+  ## Configuration options
+
+    * `:cargo` - Specify how to envoke the rust compiler. Options are:
+        - `:system` (default) - use `cargo` from the system (must by in `$PATH`)
+        - `{:rustup, <version>}` - use `rustup` to specify which channel to use.
+          Available options include: `:stable`, `:beta`, `:nightly`, or a string
+          which specifies a specific version (i.e. `"1.39.0"`).
+        - `{:bin, "/path/to/binary"}` - provide a specific path to `cargo`.
 
     * `:crate` - the name of the Rust crate (as an atom), if different from your
```
evnu: Is `as an atom` actually enforced?

rustler_mix/lib/rustler/compiler/config.ex
```diff
@@ -0,0 +1,83 @@
...
+            path: "",
+            priv_dir: "",
+            skip_compilation?: false,
+            target: nil,
+            target_dir: ""
+
+  alias Rustler.Compiler.Config
+
+  def from(otp_app, module, opts) do
+    config = Application.get_env(otp_app, module, [])
```
evnu: Should we document that configuration of a NIF is possible through the application configuration?

rustler_mix/lib/rustler/compiler/config.ex
```diff
@@ -0,0 +1,83 @@
...
+
+    cargo = Keyword.get(config, :cargo, :system)
```
evnu: With multiple NIFs present, this would allow each crate to be compiled with a different `:cargo` setting using the Application environment, right? If so, shouldn't `:cargo` be a global option?

rustler_mix/lib/rustler/compiler/config.ex
```diff
@@ -0,0 +1,83 @@
...
+    mode = config[:mode] || opts[:mode] || build_mode(Mix.env())
+    load_data = config[:load_data] || opts[:load_data] || 0
+
+    load_from =
+      case config[:load_from] do
+        {_, _} -> config[:load_from]
+        _ -> {otp_app, "priv/native/lib#{crate}"}
+      end
+
+    path = config[:path] || "native/#{crate}"
```
evnu: This is missing `|| opts[:path]`

evnu: Maybe we could simplify this function to avoid such errors:

```elixir
config = Keyword.merge(config, opts)
defaults = [default_features: true, env: [], features: [], mode: build_mode(Mix.env()), load_data: 0, ...]
merged = Keyword.merge(defaults, config)
struct!(Config, merged)
```

evnu: We could even get the defaults from the config definition itself:

```elixir
defaults = Map.from_struct(%Config{})
```

rustler_mix/lib/rustler/compiler/config.ex
```diff
@@ -0,0 +1,83 @@
...
+    target = config[:target]
+    external_resources = external_resources(path)
+    target_dir = Keyword.get(config, :target_dir, Application.app_dir(otp_app, "native/#{crate}"))
```
evnu: The `target_dir` can't be configured with `opts`.

rustler_mix/lib/rustler/compiler/config.ex
scrogson: Perhaps, if this isn't clear enough: https://github.com/rusterlium/rustler/pull/275/files#diff-fe83ab22d5a970ab547c86273d644238R13-R17

rustler_mix/lib/rustler/compiler/config.ex
scrogson: I'm not sure. I personally don't use this option and it has been here since before my time. I honestly would prefer that this wasn't an option (more for us to maintain) :smile:.

rustler_mix/lib/rustler/compiler/config.ex
scrogson: the `opts` is for users of your NIF lib to be able to configure various compile-time aspects of things...I can't think of a good reason why anyone would need to point the path to somewhere else, do you?

rustler_mix/lib/rustler/compiler.ex
scrogson: @filmor is this something we want to do here in this PR or can it wait until after we incorporate with `rebar3_cargo`?

rustler_mix/lib/rustler/compiler/config.ex
```diff
@@ -0,0 +1,87 @@
...
+      |> Keyword.get(:path)
+      |> external_resources()
+
+    opts = Keyword.put(opts, :external_resources, resources)
+
+    struct!(Config, opts)
+  end
+
+  defp external_resources(crate_path) do
+    "#{crate_path}/**/*"
```
evnu: Should this be more restrictive, e.g. `#{crate_path}/**/*.{rs,toml}`?

rustler_mix/lib/rustler.ex
```diff
@@ -16,36 +16,99 @@ defmodule Rustler do
...
     * `:load_data` - Any valid term. This value is passed into the NIF when it is
       loaded (default: `0`)
 
-  Either of the above options can be passed directly into the `use` macro like so:
+    * `:load_from` - This option allows control over where the final artifact should be
+      loaded from at runtime. By default the compiled artifact is loaded from the
+      owning `:otp_app`'s `priv/native` directory. This option comes in handy in
+      combination with the `:skip_compilation?` option in order to load pre-compiled
+      artifacts. To override the default behaviour specify a tuple:
+      `{:my_app, "priv/native/<artifact>"}`. Due to the way `:erlang.load_nif/2`
```
evnu: Does `:my_app` refer to `:otp_app` or `:crate`? Maybe use the same example as above.

rustler_mix/README.md
```diff
@@ -15,65 +14,22 @@ This package is available on [`hex.pm`](https://hex.pm/packages/rustler). To ins
...
-5. Load the NIF in your program. [See below](#loading-the-nif).
+1. Run `mix rustler.new` and follow the instructions to generate the boilerplate for your NIF.
+1. Load the NIF in your program. [See below](#loading-the-nif).
 
 ## Crate configuration
 
-The `rustler_crates` configuration is a keyword list mapping the crate name (an atom) to the NIF configuration (another keyword list).
-The NIF configuration may contain the following entries:
+The Rust crate compilation can be controlled via Mix compile-time configuration in `config/config.exs`.
+See [configuration options](https://hexdocs.pm/rustler/Rustler.html#module-configuration-options) for more
```
evnu: The anchor is broken.

rustler_mix/lib/rustler.ex
scrogson: `:otp_app`

rustler_mix/lib/rustler/compiler/config.ex
scrogson: Yeah, perhaps.

rustler_mix/README.md
scrogson: Yeah, it will be until this version is published.

PR #279, opened 2019-11-12
Validate NIF attributes

evnu:
An unknown NIF attribute should result in an error instead of being ignored. Example:

```rust
#[rustler::nif(schedul = "DirtyIo")]
pub fn dirty_io() -> Atom {
    let duration = Duration::from_millis(100);
    std::thread::sleep(duration);

    atoms::ok()
}
```

Resulting error:

```
error: custom attribute panicked
  --> rustler_tests/native/rustler_test/src/test_dirty.rs:18:1
   |
18 | #[rustler::nif(schedul = "DirtyIo")]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: message: Unknown attribute 'schedul'. Allowed attributes: ["schedule", "name"]
```

### DIFF
--- a/rustler_codegen/src/nif.rs
+++ b/rustler_codegen/src/nif.rs
@@ -7,6 +7,9 @@ pub fn transcoder_decorator(args: syn::AttributeArgs, fun: syn::ItemFn) -> Token
     let sig = &fun.sig;
     let name = &sig.ident;
     let inputs = &sig.inputs;
+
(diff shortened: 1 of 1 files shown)


scrogson:
We should also look into https://docs.rs/proc-macro-error/0.4.3/proc_macro_error/ until diagnostics become stable.

PR #280, opened 2019-11-12
Tag releases and make release script more resilient

evnu:
This changes the `release.sh` script to create releases.

* The release commit is tagged
* The `rustler_mix` documentation points to that tag
* If an error occurs during verification by compile, the commit is reverted. This is not done when publishing is already underway!

The PR adds two environment variables for the script to simplify testing:

* `DRYRUN`: Do not publish
* `DONTREVERT`: Do not revert the changes if an error happened or if `DRYRUN` was used

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -7,6 +7,7 @@
 # ## Environment Variables
 #
 # * DRYRUN: Package release, but do not publish
+# * DONTREVERT: Do not revert on error
 #
 set -e
 
@@ -19,8 +20,6 @@ if ! [[ $VERSION =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
 fi
 
 TAG="rustler-$VERSION"
-
-# FIXME get git version to revert to
 
 # Check version unpublished
 #CRATES_RET=`curl "https://crates.io/api/v1/crates/rustler/$VERSION/dependencies"`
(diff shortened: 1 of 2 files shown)


evnu:
As of 06e7f96, this does not yet work, as cargo fails to select a matching version of `rustler_codegen` for the `rustler` crate:

```
$ DRYRUN=1 ./release.sh 1.2.3
Bumping versions..
Committing version..
[tag-on-release 72a39f5] (release) 1.2.3
 5 files changed, 6 insertions(+), 6 deletions(-)
Tagging version..
~/tools/rustler/rustler ~/tools/rustler
   Packaging rustler v1.2.3 (/home/mo/tools/rustler/rustler)
   Verifying rustler v1.2.3 (/home/mo/tools/rustler/rustler)
    Updating crates.io index
error: failed to verify package tarball

Caused by:
  failed to select a version for the requirement `rustler_codegen = "^1.2.3"`
  candidate versions found which didn't match: 0.21.0, 0.20.0, 0.19.0, ...
  location searched: crates.io index
required by package `rustler v1.2.3 (/home/mo/tools/rustler/target/package/rustler-1.2.3)`
Reverting changes..
Deleted tag 'rustler-1.2.3' (was 72a39f5)
HEAD is now at 5d952d6 allow dirty
[2]    48994 exit 101   DRYRUN=1 ./release.sh 1.2.3
```

filmor:
Maybe dropping the separate `cargo package` step and publishing `rustler_codegen` first helps?

evnu:
> Maybe dropping the separate cargo package step and publishing rustler_codegen first helps?

Then, we might publish some of the crates before a later crate fails to be packaged. I think the extra step is useful, but I don't know yet how to resolve the problem.

evnu:
I think this could be the problem when packaging ([docs](https://doc.rust-lang.org/cargo/commands/cargo-package.html)):

> Cargo will ignore the path key for dependencies in published packages.

`rustler_codegen` is published already, so the path key is ignored.

evnu:
> Maybe dropping the separate cargo package step and publishing rustler_codegen first helps?

I applied that suggestion to `release.sh`.

PR #281, opened 2019-11-13
Add options to implement NIF unload and upgrade

filmor:
- [ ] Add docs (in particular about having to re-run the resource type registration)
- [ ] Clarify safety of `unload`
- [ ] Add test


drunkirishcoder:
any plan to support this?

PR #282, opened 2019-11-14
Find other cargo config files for macOS hack

### LINKED ISSUE #235: macOS platform hack does not pick up workspace cargo configuration
The platform hack for [macOS](https://github.com/rusterlium/rustler/blob/master/rustler_mix/lib/mix/tasks/compile.rustler.ex#L98) does not consider a cargo configuration within a workspace, but only the configuration local to a crate. This results in unnecessary warnings when the correct arguments are already provided within the workspace-wide configuration.

evnu:
This PR extends the macOS platform hack to work with workspaces as well. The general idea is to retrieve the workspace root from `cargo metadata` (which also works for single crates without a workspace), and then iterate over all possible configuration locations between a crate path and the workspace root. If any such potential configuration file exists, we check if it contains a target configuration specific to macOS. 

I tested this against https://github.com/evnu/NifIo/tree/macos-platform-hack by replacing the match for `:darwin` in the `make_platform_hacks/4` clause with `_`.

Fix #235

### DIFF
--- a/rustler_mix/lib/mix/tasks/compile.rustler.ex
+++ b/rustler_mix/lib/mix/tasks/compile.rustler.ex
@@ -117,11 +117,47 @@ defmodule Mix.Tasks.Compile.Rustler do
   end
 
   defp make_platform_hacks(args, crate_path, :lib, {:unix, :darwin}) do
-    path = Path.join([crate_path, ".cargo", "config"])
+    cargo_metadata =
+      case System.cmd("cargo", ["metadata", "--format-version", "1"], cd: crate_path) do
+        {cargo_metadata, 0} ->
+          cargo_metadata
 
-    if File.exists?(path) do
+        _ ->
+          raise "Failed to run 'cargo metadata --format-version 1'"
+      end
+
+    workspace_root =
+      case Regex.run(~r/"workspace_root":"(.*)"/, cargo_metadata, capture: :all_but_first) do
+        [path] ->
+          path
+
+        nil ->
+          raise "Could not find workspace_root in cargo metadata"
+      end
+
+    if !String.starts_with?(crate_path, workspace_root) do
+      raise "Expected crate_path (#{crate_path} to be within workspace_root (#{workspace_root})."
+    end
+
+    # We try to find a .cargo config up to the workspace root. The method here tries to follow
+    # the mechanism cargo uses to find configuration files. For more information, refer to the
+    # cargo documentation: https://doc.rust-lang.org/cargo/reference/config.html
+    cargo_config_exists? =
+      crate_path
+      |> Stream.iterate(&Path.dirname(&1))
+      |> Enum.take_while(fn p -> p != Path.dirname(workspace_root) end)
+      |> Enum.any?(fn p ->
+        file = Path.join([p, ".cargo", "config"])
+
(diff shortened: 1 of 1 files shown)


filmor:
I'm not a fan of regexing JSON. What is the big issue with actually pulling in a simple parser?

evnu:
> What is the big issue with actually pulling in a simple parser?

I wouldn't say it is a big issue. But we would currently only use it at one place for a very platform-specific hack. If we add a JSON library, all users would have that as a dependency as well, which seems unnecessary if we don't need that library anywhere else. @scrogson what do you think?

filmor:
Well, you can also use it to capture build-artifacts in a more reliable manner ;)

Also, would users even notice? Not sure how this works for mix, but in rebar3 plugins are not deployed along with the application but only exist during build-time.

scrogson:
I'm not a super big fan of this approach. This warning was meant to be a stop-gap until we get 1.0 out. I think the `rustler.new` task should generate the `.cargo/config` file in the root of the mix project along with a `Cargo.toml` with workspace config pointing to the NIF crate in `priv/native`.

Also...I think this here is probably all we need? https://github.com/rusterlium/rustler/blob/66616dbcf136026c1442d38d321866e0f404deff/rustler_mix/lib/rustler/compiler.ex#L83-L88

This will print the warning only if neither of those files exist regardless of the contents.

evnu:
> Also, would users even notice?

That probably depends on the users :)

> Also...I think this here is probably all we need? 

This depends how nested the directory structure is. Cargo can find the configuration in a lot of different places. Of course, we could simply assume that it should be in one of those two places instead. In my projects, I only place `.cargo` at the workspace root or within the crates itself, so this would be fine with me.

> This will print the warning only if neither of those files exist regardless of the contents.

I originally had that as well, but thought that it might be useful to take a last peek into the actual files. But I guess we can also assume that a user knowledgeable enough to create such a file in the first place should be able to also add the entry for macOS without being explicitly told here to do that.

evnu:
> This depends how nested the directory structure is.

After thinking a little bit more about this, I wonder how umbrella projects combined with cargo workspaces work if we check only two specific locations. Specifically, consider a directory structure like this:

```
/tmp/test_project tree -d -I "deps|_build"
.
├── apps
│   └── some_application
│       ├── lib
│       ├── native
│       │   └── testproject
│       │       └── src
│       └── test
└── config
```

Now, if the workspace `Cargo.toml` is defined in `.` (the root directory of the umbrella) and we place the cargo workspace configuration in `.`, then `mix compile` should be able to pick up the configuration if called in `.` as well as in `apps/testproject`. I will have to test if this works with https://github.com/rusterlium/rustler/pull/282#issuecomment-554338586.

evnu:
I just tested a simple example like described [here](https://github.com/rusterlium/rustler/pull/282#issuecomment-554925772) locally. If I place the `.cargo/config` file into the root directory of the umbrella, checking only the two locations as described [here](https://github.com/rusterlium/rustler/pull/282#issuecomment-554338586) does not find the configuration.

evnu:
Closing, we decided to drop the warning and search for the configuration completely and simply always add the link arguments on macOS.

PR #291, opened 2019-12-11
Fix Segfault in OwnedEnv::send_and_clear

scrogson:
This change fixes a segfault we found today. This was the diff that @evnu posted which does indeed fix the segfault.

I built `OTP-22.2` with `valgrind` and ran it before and after, which can be seen [here].

[here]: https://gist.github.com/scrogson/d4e11b23b946ef9fa2c7e09a4b38709d

### DIFF
--- a/rustler/src/env.rs
+++ b/rustler/src/env.rs
@@ -178,11 +178,11 @@ impl OwnedEnv {
 
         let message = self.run(|env| closure(env).as_c_arg());
 
-        let c_env = *self.env;
-        self.env = Arc::new(c_env); // invalidate SavedTerms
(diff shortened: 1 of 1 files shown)


evnu:
This looks good to me. We were missing a call to `enif_clear_env` before here, which is now done. From [the documentation](http://erlang.org/doc/man/erl_nif.html#enif_send):

> The message environment msg_env with all its terms (including msg) is invalidated by a successful call to enif_send. The environment is to either be freed with enif_free_env or cleared for reuse with enif_clear_env. An unsuccessful call will leave msg and msg_env still valid.

PR #292, opened 2019-12-11
Rename Pid to LocalPid

filmor:


hansihe:
Does it have a type alias for backwards compatibility?

scrogson:
I don't know how I feel about this one. We definitely need to add a type alias in order to not break existing projects. What is the behaviour when using `Env::send` or `OwnedEnv::send_and_clear` with remote pids? Does it crash?

If we can't prevent a remote pid from being passed into a NIF, I don't think renaming it to `LocalPid` makes much difference here.

filmor:
I'll add the type alias. You can't create an object of (C) type `ErlNifPid` that refers to a remote pid, cf http://erlang.org/doc/man/erl_nif.html#ErlNifPid. You can have a term that points to a remote pid, but you can't do much with it apart from passing it around in NIFs.

evnu:
> If we can't prevent a remote pid from being passed into a NIF, I don't think renaming it to LocalPid makes much difference here.

I think renaming helps as it indicates clearly that the PID type is restrictive. I once tried using a remote pid and was confused until I understood that decoding a remote PID is not possible. This would not have happened if the type was named `LocalPid`.

filmor:
It would be really nice if we merged this before the next release, the sooner we get the deprecation warning in the better.

PR #295, opened 2020-01-13
Drop dependency on 'which' crate

evnu:
std::process::Command determines the path to an executable using the os-specific PATH variable. As we are only interested in calling the `erl` executable when building, but not where the executable is placed within the file system, we can attempt to call it directly. The dependency was introduced with #178.


evnu:
@hansihe I assigned you as a reviewer because you reviewed #178. Feel free to reassign!

PR #300, opened 2020-01-28
Speed up generated decoder compilation times

### LINKED ISSUE #299: Compile time scales non-linearly with number of fields in structs
We are experiencing very long compile times for our crates when using `rustler`with `--release` mode. It appears that the compile time does scale superlinearly with the number of fields when using derives such as `NifStruct`. I implemented an example [here](https://github.com/evnu/rusty/tree/master/native/benchmark_compile). For the benchmark, I created some simple crates which contain only one struct using `NifStruct` and a number of simple fields.

Example run:

```
~/tools/rusty/native/benchmark_compile(master ✗) ./benchmark.sh
number_of_fields  time
0 0.03
1 0.24
2 0.27
4 0.54
8 1.76
16 9.11
32 60.81
48 156.27
64 407.51
~/tools/rusty/native/benchmark_compile(master ✗) ./plot.gpi
```

plot

For comparision, here are the results when using the debug mode:

plot

As expected, the compile time is much faster. It also appears to be more linear than with release mode. I suspect that we somehow make the life of LLVM harder w.r.t. the generated code.

evnu:
This pull request speeds up compilation by extracting common code from field assignments into helper functions, and by preferring local variables for initial assignment instead of assigning struct fields directly. See #299 for the corresponding issue.

Fix #299.

### DIFF
--- a/rustler_codegen/src/ex_struct.rs
+++ b/rustler_codegen/src/ex_struct.rs
@@ -55,6 +55,7 @@ pub fn transcoder_decorator(ast: &syn::DeriveInput) -> TokenStream {
 fn gen_decoder(ctx: &Context, fields: &[&Field], atoms_module_name: &Ident) -> TokenStream {
     let struct_type = &ctx.ident_with_lifetime;
     let struct_name = ctx.ident;
+    let struct_name_str = struct_name.to_string();
 
     let field_defs: Vec<TokenStream> = fields
         .iter()
@@ -62,16 +63,8 @@ fn gen_decoder(ctx: &Context, fields: &[&Field], atoms_module_name: &Ident) -> T
             let ident = field.ident.as_ref().unwrap();
             let ident_str = ident.to_string();
             let atom_fun = Ident::new(&format!("atom_{}", ident_str), Span::call_site());
-            let error_message = format!(
-                "Could not decode field :{} on %{}{{}}",
-                ident.to_string(),
-                struct_name.to_string()
-            );
             quote! {
-                #ident: match ::rustler::Decoder::decode(term.map_get(#atom_fun().encode(env))?) {
-                    Err(_) => return Err(::rustler::Error::RaiseTerm(Box::new(#error_message))),
-                    Ok(value) => value
-                }
+                #ident: try_decode_field(env, term, #atom_fun())?
             }
         })
         .collect();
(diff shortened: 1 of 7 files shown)


evnu:
Some benchmark results from running `meta_benchmark.sh` in https://github.com/evnu/rusty/tree/master/native/benchmark_compile.

```
~/tools/rusty/native/benchmark_compile(master ✗) ./meta_benchmark.sh
```

|Derive Type|master|299-speed-up-decoder-compilation|
|-|-|-|
|NifMap|NifMap_master|NifMap_299-speed-up-decoder-compilation|
|NifStruct|NifStruct_master|NifStruct_299-speed-up-decoder-compilation|
|NifRecord|NifRecord_master|NifRecord_299-speed-up-decoder-compilation|
|NifTuple|NifTuple_master|NifTuple_299-speed-up-decoder-compilation|

This looks very promising. I was not yet able to refactor `NifRecord` and `NifTuple` to achieve a similar speed up. The problem is that we need to introduce `let` assignments for the fields, but with tuple structs we do not have a field name available to name the let variable.

evnu:
I managed to use `let` for tuple and records as well and squashed it into 2c36dee. Now, compilation is fast for `NifTuple` and `NifRecord` as well:

|Derive Type|master|299-speed-up-decoder-compilation|
|-|-|-|
|NifRecord|NifRecord_master|NifRecord_299-speed-up-decoder-compilation|
|NifTuple|NifTuple_master|NifTuple_299-speed-up-decoder-compilation|

rustler_codegen/src/ex_struct.rs
```diff
@@ -55,23 +55,20 @@ pub fn transcoder_decorator(ast: &syn::DeriveInput) -> TokenStream {
...
-                "Could not decode field :{} on %{}{{}}",
-                ident.to_string(),
-                struct_name.to_string()
-            );
             quote! {
-                #ident: match ::rustler::Decoder::decode(term.map_get(#atom_fun().encode(env))?) {
-                    Err(_) => return Err(::rustler::Error::RaiseTerm(Box::new(#error_message))),
-                    Ok(value) => value
-                }
+                let #ident = try_decode_field(env, term, #atom_fun())?;
```
hansihe: Just so things are a bit more robust, maybe we should be using a number index or something as the variable name in the generated code?

rustler_codegen/src/map.rs
```diff
@@ -75,7 +75,28 @@ fn gen_decoder(ctx: &Context, fields: &[&Field], atoms_module_name: &Ident) -> T
                 use #atoms_module_name::*;
 
                 let env = term.get_env();
-                Ok(#struct_name { #(#field_defs),* })
+
+                fn try_decode_field<'a, T>(
```
hansihe: :+1:

rustler_codegen/src/ex_struct.rs
evnu: 01d749c. Note that I am unsure if we actually need to do this, as we have the complete function under our control here (we generate it right now). In any case, it shouldn't hurt.

evnu: This conflicts with #303. When a field is named `r#override`, the generated name is something like `RUSTLER_struct_field_0_r#override`, which isn't a valid identifier.

evnu: I extracted generating the identifiers to the `Context` module in 95520dd.

PR #302, opened 2020-02-06
Common cargo

filmor:
Use https://github.com/rusterlium/erlang-cargo/ to have a common understanding with https://github.com/rusterlium/rebar3_cargo/ about how to compile and where to put the results.


filmor:
I will clean this up further (and will publish `erlang-cargo` to Hex for use in here), but this is finally passing all tests, would be nice if you could review and also check in your own applications, whether this works for you.

rustler_mix/lib/mix/tasks/rustler.new.ex
```diff
@@ -24,7 +24,6 @@ defmodule Mix.Tasks.Rustler.New do
 
   for {format, source, _} <- @basic do
     unless format == :keep do
-      @external_resource Path.join(root, source)
```
filmor, line 27: This and the other occurrences of `@external_resource` will be readded.

rustler_mix/lib/rustler.ex
```diff
@@ -29,13 +29,8 @@ defmodule Rustler do
       value. If you have more than one crate in your project, you will need to
       be explicit about which crate you intend to use.
 
-    * `:default_features` - a boolean to specify whether the crate's default features
-      should be used.
-
     * `:env` - Specify a list of environment variables when envoking the compiler.
 
-    * `:features` - a list of features to enable when compiling the crate.
-
```
filmor, line 38: I can probably readd this if required.

rustler_mix/lib/rustler.ex
```diff
@@ -52,12 +47,6 @@ defmodule Rustler do
       - When `Mix.env()` is `:dev` or `:test`, the crate will be compiled in `:debug` mode.
       - When `Mix.env()` is `:prod` or `:bench`, the crate will be compiled in `:release` mode.
 
-    * `:path` - By default, rustler expects the crate to be found in `native/<crate>` in the
-      root of the project. Use this option to override this.
-
```
filmor, line 57: This is not necessary anymore, as we now expect a `Cargo.toml` in the root of the project.

.github/workflows/main.yml
```diff
@@ -135,13 +137,6 @@ jobs:
           - erlang: 21.x
             elixir: 1.6
 
-          - erlang: 20.3.8.23
-            elixir: 1.8
```
evnu, line 139: I think it might surprise users that they must run OTP >= 21 with 1.6 <= Elixir <= 1.9. The [Compatibility between Elixir and Erlang/OTP ](https://hexdocs.pm/elixir/compatibility-and-deprecations.html#compatibility-between-elixir-and-erlang-otp) claims that Elixir 1.6 through 1.9 support OTP 20.

filmor: I can check again, which >otp20 API I used, I doubt it was important.

filmor: Hah, whatever the actual problem here was, it's not there anymore. I've set up CI builds and tests for `erlang-cargo` as well for OTP 20-23 and they passed, so I removed this commit.

rustler_mix/lib/mix/tasks/rustler.new.ex
filmor: Hmm, actually I don't like this, it would mean (if I understand things correctly) that automatic recompilation is only triggered if `lib.rs` is changed, not if any of its dependencies (like other `.rs` files or a crate dependencies) are changed, is that right?

filmor: Yes, I'll do this via `__mix_recompile__?` instead.

.github/workflows/main.yml
evnu: :+1:

rustler_mix/lib/rustler/compiler/config.ex
```diff
@@ -2,31 +2,20 @@ defmodule Rustler.Compiler.Config do
...
-            crate: nil,
-            default_features: true,
-            env: [],
-            external_resources: [],
-            features: [],
+  defstruct crate: nil,
             lib: true,
             load_data: 0,
-            load_from: nil,
+            load_path: nil,
```
scrogson, line 15: The intention for this was to allow pre-compiled artifacts to be loaded from your own `priv` dir.

filmor: I know, maybe we should discuss this in a bit more detail.

rustler_mix/lib/mix/tasks/rustler.new.ex
evnu: `__mix_recompile__?` seems very new, I think we might not be able to support older versions of Elixir with that anymore?

filmor: I just found it in the sources. Worst case: It won't be better than what we have now :)

The issue is that you can't really tell `mix` up-front what exact files will influence the final object files of a Rust project. I'll think about it.

rustler_mix/lib/rustler/compiler/server.ex
```diff
@@ -0,0 +1,45 @@
+defmodule Rustler.Compiler.Server do
+  use GenServer
```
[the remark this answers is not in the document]
filmor: `Rustler.Compiler` is run in parallel by mix, it is required.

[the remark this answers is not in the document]
filmor: And actually, we do hold state, the compile artifacts.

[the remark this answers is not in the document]
filmor: To elaborate a bit: I ran into the problem of multiple parallel `cargo` invocations on the same thing in the tests, which leads to `cargo` erroring out. Since in the current form in this PR we always run `cargo build` on the whole workspace, by using this server it won't matter and we have all relevant information available to all callers. I tried hard to find a mechanism built into mix to do this but couldn't find anything.

rustler_mix/lib/rustler/compiler/server.ex
```diff
@@ -0,0 +1,45 @@
...
+
+    # This drops the unique key in favour of the crate name
+    artifacts =
+      artifacts
+      |> Map.new(&{:cargo_artifact.name(&1), &1})
+
+    {:reply, artifacts, artifacts}
+  end
+
+  def handle_call(:compile, _from, artifacts) do
```
filmor, line 42: @evnu This is the case where the state is used, if `state != nil`.

evnu: Ah, missed that, sorry.

rustler_mix/lib/rustler/compiler/server.ex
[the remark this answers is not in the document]
evnu: :+1:

rustler_mix/lib/rustler/compiler.ex
```diff
@@ -1,174 +1,80 @@
...
+        rel_filename = Path.basename(filename)
+        rel_dest = Path.relative_to_cwd(dest)
 
-  defp make_base_command({:rustup, version}) do
-    if Rustup.version() == :none do
-      throw_error(:rustup_not_installed)
-    end
+        shell.info("  Copying #{rel_filename} to #{rel_dest}")
+        File.mkdir_p!(Path.dirname(dest))
+        File.copy!(filename, dest)
```
evnu, line 39: I think we should delete the destination first, see the deleted code: https://github.com/rusterlium/rustler/pull/302/files#diff-cb1f5883582c0a6149e7a7c774f4702cR39

We found a problem due to `mmap` when simply copying, see https://github.com/rusterlium/rustler/issues/128: If another node is currently using the NIF, it has the lib mmap'ed, which results in segfaults if we overwrite it with copy.

filmor: Actually, I'd rather have this case error out entirely. You can get a new version pushed (safely) by bumping the crate version (there might be some trickyness involved with dynamic linkers that remember the filenames). On Windows you will never be able to overwrite an in-use DLL file.

filmor:
TODO:

- [ ] Add `load_from` back to side-step the whole recompilation process
- [ ] Allow overriding the default env-dependent `is_release` per module

rustler_mix/lib/rustler/compiler.ex
evnu: To give some more context: I encountered segfaults when running `mix test` and editing+compiling in two separate terminals. So this is something which we need to guard against here.

evnu:
@filmor should we pick this PR up again? Can I somehow help to move it along?

filmor:
As a first step, I'll rebase it. Maybe we could have a call on this to talk about the next steps?

evnu:
> As a first step, I'll rebase it. Maybe we could have a call on this to talk about the next steps?

Yep, lets do that.

PR #306, opened 2020-02-18
Add Elixir v1.10 to github actions

evnu:
This adds Elixir v1.10 to the CI pipeline. It seems that the setup-elixir action does not yet support this version. See [here](https://github.com/actions/setup-elixir/issues/16).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -116,6 +116,8 @@ jobs:
       matrix:
         pair:
           - erlang: 22.x
+            elixir: "1.10"
+          - erlang: 22.x
             elixir: 1.9
           - erlang: 22.x
             elixir: 1.8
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -115,6 +115,8 @@ jobs:
     strategy:
       matrix:
         pair:
+          - erlang: 22.x
+            elixir: 1.10
```
filmor: Use a string here, otherwise it will be `1.1` (parsed as a float).

evnu: Thanks, will give that a try!

evnu:
@filmor https://github.com/rusterlium/rustler/runs/455245770?check_suite_focus=true:

```
  ##[error]Invalid Version: 1.10
```

PR #307, opened 2020-03-03
Add support for enif_whereis_pid

evnu:
This pull request adds support for `enif_whereis_pid`, introduced with OTP 20. Currently, the implementation always requires a valid `Env`. The safety section of the unsafe wrapper indicates that `enif_whereis_pid` can also be called with `env == null`, iff the caller is a created process. With the current implementation, this cannot be triggered, as an `Env` must always be passed into the safe wrapper.

### DIFF
--- a/rustler/src/types/local_pid.rs
+++ b/rustler/src/types/local_pid.rs
@@ -1,5 +1,5 @@
 use crate::wrapper::{pid, ErlNifPid};
-use crate::{Decoder, Encoder, Env, Error, NifResult, Term};
+use crate::{Decoder, Encoder, Env, Error, NifResult, Term, Atom};
 use std::mem::MaybeUninit;
 
 #[derive(Clone)]
@@ -10,6 +10,13 @@ pub struct LocalPid {
 impl LocalPid {
     pub fn as_c_arg(&self) -> &ErlNifPid {
         &self.c
+    }
+
+    ///
+    /// Look up a local process by its registered name.
+    ///
+    pub fn whereis<'a>(env: Env<'a>, name: Atom) -> Option<Self> {
+        unsafe { pid::whereis(env.as_c_arg(), name.as_c_arg()).map(|pid| LocalPid { c: pid }) }
     }
 }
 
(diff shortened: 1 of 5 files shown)


rustler/src/types/local_pid.rs
```diff
@@ -11,6 +11,13 @@ impl LocalPid {
     pub fn as_c_arg(&self) -> &ErlNifPid {
         &self.c
     }
+
+    ///
+    /// Look up a local process by its registered name.
+    ///
+    pub fn whereis<'a>(env: Env<'a>, name: Atom) -> Option<Self> {
```
evnu: This might as well be a function on `Env` instead. I am not sure what we should prefer here.

```rust
env.whereis(name)
```

vs

```rust
LocalPid::whereis(env, name)
```

rustler/src/wrapper/pid.rs
```diff
@@ -16,3 +16,20 @@ pub unsafe fn get_local_pid(env: NIF_ENV, term: NIF_TERM) -> Option<ErlNifPid> {
 pub unsafe fn make_pid(env: NIF_ENV, pid: ErlNifPid) -> NIF_TERM {
     rustler_sys::enif_make_pid(env, pid)
 }
+
+///
+/// Look up a local process by its registered name.
+///
+/// # Safety
+///
+/// `env` must only be null if the calling process is a created process. Otherwise,
```
evnu: The [section](http://erlang.org/doc/man/erl_nif.html#enif_whereis_pid) in the documentation reads:

> The environment of the calling process. Must be NULL only if calling from a created thread.

Does that mean that it **MUST** be `NULL` if calling from a created thread, or that it **MAY** be `NULL` in that case?

hansihe: I created an issue on the erlang issue tracker for this.

https://bugs.erlang.org/browse/ERL-1399

I think they mean it in the strictest sense, but would be nice to get confirmation about this.

hansihe: > It's the first: NULL for created thread, otherwise the environment argument (process bound environment or callback environment).

PR #311, opened 2020-05-11
Add AsRef for Binary

vihu:
This is mostly just for convenience.

### DIFF
--- a/rustler/src/types/binary.rs
+++ b/rustler/src/types/binary.rs
@@ -239,6 +239,12 @@ impl<'a> Encoder for Binary<'a> {
     }
 }
 
+impl<'a> AsRef<[u8]> for Binary<'a> {
+    fn as_ref(&self) -> &[u8] {
+        &self.as_slice()
+    }
+}
+
 /// ## Binary terms
 impl<'a> Term<'a> {
     pub fn into_binary(self) -> NifResult<Binary<'a>> {


evnu:
Thanks for the pull request! Out of curiosity, when is `.as_ref()` preferable over the existing `.as_slice()` function?

vihu:
> Thanks for the pull request! Out of curiosity, when is `.as_ref()` preferable over the existing `.as_slice()` function?

I believe `as_slice` doesn't let you do polymorphism, i.e. you cannot specify an input type parameter. I know I'm pretty much using the same definition for the `as_ref` here, the difference being the input type. I _think_ this behaves a bit more appropriately when inter-oping with erlang refs. I'll try n find some example code where I hit rustler complaining that `AsRef` wasn't implemented for `Binary`, hence this draft :)

evnu:
Hey @vihu, did you manage to find an example of rustler complaining about `AsRef`?

vihu:
> Hey @vihu, did you manage to find an example of rustler complaining about `AsRef`?

Oh man, this was a while ago, I don't quite recall where I saw it. I think it was somewhere in [rocker](https://github.com/Vonmo/rocker) when I was trying to update it to use new ruslter api.

In any case, I think this PR is probably stale now, so I'm fine if you just want to close it.

evnu:
@vihu Thanks for the reply! I will close this for now. We can reopen if someone stumbles over the problem again.

PR #314, opened 2020-06-08
Implement Encode/Decode for HashMap

### LINKED ISSUE #312: Decoding Maps of arbitrary length
**Rustler Version:** 0.22.0-rc.0

This is perhaps more a question of something I am doing wrong but it doesn't seem like `NifMap` supports maps of length N > 1. For example trying to decode the following map:

```elixir
%{
    1 => %SomeStruct{},
    2 => %SomeStruct{}
   }
```

Where `%SomeStruct{}` is appropriately set up in the Nif as `NifStruct` and can be decoded singularly and the map key is just a `u32`.

What's the correct type for a map of arbitrary length? The documentation also only supplies examples for a map with 1 KV pair.

To be clear I realise I can use MapIterator when it's passed in as part of a NIF function but I am talking about maps that are part of structs, for example:

```rust
#[derive(Clone, Debug, Eq, Hash, NifMap, PartialEq)]
pub struct Counts {
    id: String,
    count: u32,
}

#[derive(Clone, Debug, Eq, Hash, NifStruct, PartialEq)]
#[module = "SomeModule.SomeStruct"]
pub struct SomeStruct {
    pub r#type: i32,
    pub remaining: u32,

    /// An elixir map `%{Counts.id() => pos_integer()}`
    pub counts: Counts,
}
```

```elixir
%SomeStruct{type: 1, remaining: 0, counts: %{"hello" => 12, "world" => 25}}
```

If I try to make `pub counts` of type `MapIterator` I have to put lifetimes everywhere, I assume because it shares the lifetime of the `Env`

evnu:
Fix #312 .

### DIFF
--- a/rustler/src/types/mod.rs
+++ b/rustler/src/types/mod.rs
@@ -133,9 +133,11 @@ where
     V: Decoder<'a>,
 {
     fn decode(term: Term<'a>) -> NifResult<Self> {
+        let size = term.map_size()?;
+
         let it = MapIterator::new(term).ok_or_else(|| Error::BadArg)?;
 
-        let mut map = std::collections::HashMap::new();
+        let mut map = std::collections::HashMap::with_capacity(size);
 
         for (k, v) in it {
             let k = k.decode()?;
--- a/rustler/src/types/mod.rs
+++ b/rustler/src/types/mod.rs
@@ -126,3 +126,37 @@ where
         }
     }
 }
+
+impl<'a, K, V> Decoder<'a> for std::collections::HashMap<K, V>
+where
+    K: Decoder<'a> + Eq + std::hash::Hash,
+    V: Decoder<'a>,
+{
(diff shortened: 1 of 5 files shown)


rustler/src/types/mod.rs
```diff
@@ -126,3 +126,37 @@ where
...
+
+impl<'a, K, V> Decoder<'a> for std::collections::HashMap<K, V>
+where
+    K: Decoder<'a> + Eq + std::hash::Hash,
+    V: Decoder<'a>,
+{
+    fn decode(term: Term<'a>) -> NifResult<Self> {
+        let it = MapIterator::new(term).ok_or_else(|| Error::BadArg)?;
+
+        let mut map = std::collections::HashMap::new();
```
evnu, line 138: It might make sense to retrieve the map's size with `enif_get_map_size` and call `with_capacity()` here.

evnu: d5a23a8

PR #320, opened 2020-06-23
Fix clippy lints

evnu:


evnu:
In b6cb60eaea39739194584977ccc3f6fc69d4482d, I disabled the warning on missing `Safety` sections in function docs. This change must be reverted.

evnu:
Rebased on master after #321.

Issue #324, opened 2020-07-02
[Release Candidate] on_load error when checking out different branches


hazardfn:
Sometimes when switching between branches, running mix deps.get and then compiling this error pops up:

```
14:14:56.816 [warn]  The on_load function for module asn1rt_nif returned:
{:badarg, [{:erlang, :load_nif, [:erlang, :apply], []}, {:asn1rt_nif, :load_nif, 0, [file: 'asn1rt_nif.erl', line: 58]}, {:code_server, :"-handle_on_load/5-fun-0-", 1, [file: 'code_server.erl', ...]}]}
14:14:56.835 [error] Process #PID<0.3742.0> raised an exception
** (ArgumentError) argument error
    :erlang.load_nif(:erlang, :apply)
    (asn1 5.0.13) asn1rt_nif.erl:58: :asn1rt_nif.load_nif/0
    (kernel 7.0) code_server.erl:1355: anonymous fn/1 in :code_server.handle_on_load/5
```

It doesn't seem to cause any immediate/noticeable issues and started happening for me when using OTP 23.0. The thing that stands out to me is the arguments to `load_nif/2` look completely wrong `(:erlang, :apply)`

The error comes from the `use Rustler` macro line hence why I am reporting it here, some changes to NIFs were bought into 23.0 but nothing that was supposed to break compatibility as far as I know. One thing that could be related is: https://github.com/erlang/otp/commit/a4e8656d1aa4bde0bbfab6a52585add5d7951bd1

evnu:
I managed to reproduce this on 387fd8a (rustler-0.22.0-rc.0) as well as on the current master 427da64. Running `mix test` in `rustler_tests`:

```
~/tools/rustler/rustler_tests(387fd8a ✗) mix test
Compiling NIF crate :binary_example (native/binary_example)...
    Finished dev [unoptimized + debuginfo] target(s) in 0.00s
Compiling NIF crate :rustler_test (native/rustler_test)...
    Finished dev [unoptimized + debuginfo] target(s) in 0.01s
Compiling NIF crate :deprecated_macros (native/deprecated_macros)...
    Finished dev [unoptimized + debuginfo] target(s) in 0.01s

21:01:37.982 [error] Process #PID<0.242.0> raised an exception
** (ArgumentError) argument error
    :erlang.load_nif(:erlang, :apply)
    (rustler_test 0.0.1) lib/rustler_test.ex:6: RustlerTest.__init__/0
[...]

PR #325, opened 2020-07-02
ci: Test OTP 23

evnu:
@hazardfn found that NIF loading sometimes fails, see https://github.com/rusterlium/rustler/issues/324. We should also test against OTP 23.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -138,6 +138,9 @@ jobs:
     strategy:
       matrix:
         pair:
+          - erlang: "23.0.3"
+            elixir: "1.10.3"
+
           - erlang: 22.x
             elixir: "1.10.1"
           - erlang: 22.x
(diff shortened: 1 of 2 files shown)


evnu:
As #324 is caused by an upstream bug, we might have to wait until the OTP fix is released before merging this change here.

.github/workflows/main.yml
```diff
@@ -138,6 +138,9 @@ jobs:
     strategy:
       matrix:
         pair:
+          - erlang: "23.0.2"
```
evnu, line 141: I tried to use `23.x` here, but `setup-elixir` was not able to convert that into an actual version.

evnu: I changed it back to 23.x, we should somehow make this work instead of pinning the patch version.

evnu:
@filmor I added an `allow` rule for clippy, see the diff. I think we can safely ignore that lint for now.

PR #327, opened 2020-09-03
Remove OwnedBinary's release flag

JayKickliter:
`release` signals to `OwnedBinary`'s `Drop` implementation whether or not call `enif_release_binary()`. The only place that changes `release` from `true` to `false` is in `Binary::from_owned()`. We can therefore remove the release flag altogether by simply preventing `drop()` in `Binary::from_owned()`.

### DIFF
--- a/rustler/src/types/binary.rs
+++ b/rustler/src/types/binary.rs
@@ -8,17 +8,11 @@ use std::ops::{Deref, DerefMut};
 
 // Owned
 
-pub struct OwnedBinary {
-    inner: ErlNifBinary,
-    release: bool,
-}
+pub struct OwnedBinary(ErlNifBinary);
 
 impl<'a> OwnedBinary {
     pub unsafe fn from_raw(inner: ErlNifBinary) -> OwnedBinary {
-        OwnedBinary {
-            inner,
-            release: true,
-        }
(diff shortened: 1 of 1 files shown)


evnu:
@JayKickliter Thanks for the PR! I fixed the clippy lints on master. Could you rebase please?

filmor (approved):

rustler/src/types/binary.rs
```diff
@@ -132,18 +121,15 @@ pub struct Binary<'a> {
 impl<'a> Binary<'a> {
     /// Releases a given `OwnedBinary` to the VM.
     /// After this point the binary will be immutable.
-    pub fn from_owned(mut bin: OwnedBinary, env: Env<'a>) -> Self {
-        bin.release = false;
+    pub fn from_owned(bin: OwnedBinary, env: Env<'a>) -> Self {
+        let mut bin = std::mem::ManuallyDrop::new(bin);
```
filmor, line 125: It might be a good idea to add a comment here explaining that we intentionally skip the `drop()` at the end of scope.

JayKickliter: good suggestion, will add.

filmor:
Perfect, thanks :)

PR #330, opened 2020-09-10
Improve binary documentation

JayKickliter:
For my colleagues and I, one of the hardest parts of coming up to speed with Rustler has been dealing with binaries. Even those of us with years of Rust experience were confused by the difference between `Binary` and `OwnedBinary`, despite the fact that `Owned` should say it all.  I've tried my best to capture my confusion before I'm too familiar with codebase. 

Included in this PR:

- add module-level for the binary module, thus I had to remove the `doc_hidden` tag
- When adding a `# Panics` section for `OwnedBinary::from_unowned`, I noticed that it could be simplified to the point that it's obvious that it will not panic (it calls `copy_from_slice` internally, which can panic, but won't since lengths will be checked before that call)
- documents all binary api functions, even if trivially, for completeness sake 

For reviewers: please run `cargo doc --open` and try all the intra-doc links in the `binary` module if possible. I've `rustdoc` to be very unreliable at warning on broken links.

### DIFF
--- a/rustler/src/types/binary.rs
+++ b/rustler/src/types/binary.rs
@@ -7,7 +7,7 @@
 //! The _owned_ in `OwnedBinary` refers to the fact that it owns the binary it
 //! wraps. The _owner_ of an `OwnedBinary` is free to modify its contents. Ownership
 //! lasts until it is dropped or consumed by converting it into a regular
-//! `Binary`. An `OwnedBinary` can not be copied or cloned and is thus always moved.
+//! `Binary`. An `OwnedBinary` cannot be copied or cloned and is thus always moved.
 //!
 //! The `Binary` type is an immutable shared-reference to a binary. `Binary`s are
 //! cheap to copy: all copies of a `Binary` point to the original `Binary`'s
--- a/rustler/src/types/binary.rs
+++ b/rustler/src/types/binary.rs
@@ -1,13 +1,97 @@
-use crate::wrapper::binary::{alloc, realloc, ErlNifBinary};
-use crate::{Decoder, Encoder, Env, Error, NifResult, Term};
+//! Safe wrappers around Erlang binaries.
+//!
+//! Rustler provides two binary types: [`Binary`] and [`OwnedBinary`].  Both
+//! represent a contiguous region `u8`s, and they both use the Erlang allocator. The
+//! primary difference between the two is their ownership semantics.
+//!
+//! The _owned_ in `OwnedBinary` refers to the fact that it owns the binary it
+//! wraps. The _owner_ of an `OwnedBinary` is free to modify its contents. Ownership
+//! lasts until it is dropped or consumed by converting it into a regular
+//! `Binary`. An `OwnedBinary` can not be copied or cloned and is thus always moved.
+//!
+//! The `Binary` type is an immutable shared-reference to a binary. `Binary`s are
+//! cheap to copy: all copies of a `Binary` point to the original `Binary`'s
+//! data. Additionally, a `Binary`'s lifetime is tied to that of the NIF's [`Env`],
+//! preventing outstanding references to the data after a NIF returns.
+//!
+//! # Examples
+//!
+//! Constructing an `OwnedBinary`:
+//!
+//! ```no_run
+//! # use rustler::OwnedBinary;
+//! {
+//!     let mut bin = OwnedBinary::new(5).expect("allocation failed");
+//!     bin.as_mut_slice().copy_from_slice("hello".as_bytes());
+//! } // <- `bin` is dropped here
+//! ```
+//!
+//! The following NIF takes a binary as its only parameter and returns a new binary
+//! where each element is exclusive-or'ed with a constant:
+//!
+//! ```no_run
+//! # use rustler::{Env, OwnedBinary, Binary, NifResult, Error};
+//! #[rustler::nif]
+//! fn xor_example<'a>(env: Env<'a>, bin: Binary<'a>) -> NifResult<Binary<'a>> {
+//!     let mut owned: OwnedBinary = bin.to_owned().ok_or(Error::Term(Box::new("no mem")))?;
+//!     for byte in owned.as_mut_slice() {
+//!         *byte ^= 0xAA;
+//!     }
+//!
+//!     // Ownership of `owned`'s data is transferred to `env` on the
+//!     // following line, so no additional heap allocations are incurred.
+//!     Ok(Binary::from_owned(owned, env))
+//! }
+//! ```
+//!
+//! The contents of a newly-allocated `OwnedBinary` is not initialized to any
+//! particular value. If your usage of the binary requires the it's data to be
+//! zeroed, for example, then you must explicit zero it. In this example, we
+//! manually zeroize the binary before passing it as slice to a third party
+//! function.
+//!
+//! ```no_run
+//! # fn some_third_party_api(buf: &mut [u8]) {
+//! #     for elem in buf {
+//! #         if *elem == 0 { *elem = 1 } else { panic!("Not a zero!") }
+//! #     }
+//! # }
+//! # use rustler::{Env, OwnedBinary, Binary, NifResult, Error};
+//! #[rustler::nif]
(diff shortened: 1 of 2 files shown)


rustler/src/types/binary.rs
```diff
@@ -1,43 +1,105 @@
...
+//! the Erlang allocator. The primary difference between the two is ownership
+//! semantics.
+//!
+//! The _owned_ in `OwnedBinary` refers to the fact that it owns the VM binary is
+//! wraps, and the _owner_ of an `OwnedBinary` is free to modify its contents
+//! without fear of access violations or upsetting the VM. Ownership lasts until it
+//! is dropped by either falling out of scope, or converting it into a regular
+//! `Binary` via [`to_owned`] or one of the conversion traits. Once dropped, the VM
+//! is allowed to the free the backing memory at its leisure. Similar to `Vec`,
+//! `OwnedBinary` does not implement `copy` [^1], so passing one around by value is
```
filmor, line 14: Nit: In Rust, if something implements `Copy`, the assumption is usually that it can by efficiently passed by value. Since `Vec`s can be `clone`d, I guess it would be better to leave out that comparison and just say

     An `OwnedBinary` can not be copied or cloned and is thus always moved.

rustler/src/types/binary.rs
```diff
@@ -1,43 +1,105 @@
...
+//!
+//! Rustler provides two types for wrapping Erlang binaries: [`Binary`] and
+//! [`OwnedBinary`].  Both represent a contiguous region `u8`s, and they both use
+//! the Erlang allocator. The primary difference between the two is ownership
+//! semantics.
+//!
+//! The _owned_ in `OwnedBinary` refers to the fact that it owns the VM binary is
+//! wraps, and the _owner_ of an `OwnedBinary` is free to modify its contents
+//! without fear of access violations or upsetting the VM. Ownership lasts until it
+//! is dropped by either falling out of scope, or converting it into a regular
```
filmor, line 11: Ownership lasts until it is dropped or consumed by converting it into a regular `Binary`...

rustler/src/types/binary.rs
```diff
@@ -1,43 +1,105 @@
...
+//! `Binary` via [`to_owned`] or one of the conversion traits. Once dropped, the VM
+//! is allowed to the free the backing memory at its leisure. Similar to `Vec`,
```
filmor, line 13: This is not quite true.

1. I'd use "dropped" only for `Drop::drop` (as commented above)
2. While the NIF is still executing, an owned binary that was converted to a `Binary` does not free its backing store, it will still be a the exact same place but is considered read-only

I think I'd just remove that line entirely, the semantics of the types capture this nicely. Maybe replace it by something like

    A `Binary` belongs to a particular `Env`, so it can only be used for the duration of the remaining NIF call.

rustler/src/types/binary.rs
```diff
@@ -1,43 +1,105 @@
...
+//! `OwnedBinary` does not implement `copy` [^1], so passing one around by value is
+//! cheap and does not incur any additional heap allocations.
+//!
+//! The `Binary` type is an immutable shared-reference to a VM binary. `Binary`s are
+//! cheap to copy: all copies of of `Binary` point to the original `Binary`'s
+//! data. Additionally, a `Binary`'s lifetime is tied to the lifetime of a VM
```
filmor, line 19: The `Env` is not a VM. That makes it sound like you will get the same `Env` on every NIF call, which is not the case. The `Binary`'s lifetime, like all other VM-owned objects, is tied to exactly one NIF execution. You are not allowed to access any of these objects once the NIF is finished.

filmor: Actually, the second part of the sentence makes that clear, but `Env != VM` is still important, I think.

filmor: Actually actually, the `Env` is always either a "process environment" (i.e. tied to one particular process) or a manually allocated one to create terms to send around.

rustler/src/types/binary.rs
```diff
@@ -1,43 +1,105 @@
...
+//!
+//! ```no_run
+//! # use rustler::OwnedBinary;
+//! {
+//!     let mut bin = OwnedBinary::new(5).expect("VM couldn't allocated requested bytes");
+//!     bin.as_mut_slice().copy_from_slice("hello".as_bytes());
+//! } // <- `bin` is dropped here, VM allowed to free data
+//! ```
+//!
+//! The following NIF takes a binary as is only parameter and returns a new binary
```
filmor, line 35: Typo: as *its*  only parameter

rustler/src/types/binary.rs
```diff
@@ -1,43 +1,105 @@
...
+//! Rustler provides two types for wrapping Erlang binaries: [`Binary`] and
+//! [`OwnedBinary`].  Both represent a contiguous region `u8`s, and they both use
+//! the Erlang allocator. The primary difference between the two is ownership
+//! semantics.
+//!
+//! The _owned_ in `OwnedBinary` refers to the fact that it owns the VM binary is
+//! wraps, and the _owner_ of an `OwnedBinary` is free to modify its contents
+//! without fear of access violations or upsetting the VM. Ownership lasts until it
+//! is dropped by either falling out of scope, or converting it into a regular
+//! `Binary` via [`to_owned`] or one of the conversion traits. Once dropped, the VM
```
filmor, line 12: `to_owned` is not the right method.

JayKickliter:
@filmor thanks for all the feedback, I'll use all your suggestions. I figured I had factual errors, hence opening an early draft PR.

rustler/src/types/binary.rs
JayKickliter: >  the assumption is usually that it can by efficiently passed by value

Yeah. I was trying to go into a little more detail than you'd normally see in rust documentation. Except for me, everyone else at my company using Rustler is an Erlang-first developer, with varying familiarity with Rust. I think that might be case for other users of this crate. 

I guess what I'm really trying to say there is "don't worry, it's cheap to move these binary types around", but I think your suggestion makes the most sense.

JayKickliter:
I think I have addressed most/all of @filmor's comments, but may have introduced more factual errors in the process. 

I've gotten a little busy lately, so I may not be able complete everything in the in checklist in the PR overview at the top, but I'll leave this PR as draft for little a longer.

JayKickliter:
I removed the `draft` tag as this PR is about as complete as I can make it. I've made made some changes since @filmor's initial review of the draft version, so I welcome any further comments.

rustler/src/types/binary.rs
```diff
@@ -1,57 +1,147 @@
...
 
 impl<'a> OwnedBinary {
     pub unsafe fn from_raw(inner: ErlNifBinary) -> OwnedBinary {
         OwnedBinary(inner)
     }
 
-    /// Allocates a new OwnedBinary with size `size`.
+    /// Allocates a new `OwnedBinary` with size `size`.
+    ///
+    /// Memory is not be initialized. If uninitialized memory is undesirable, set it
```
evnu, line 104:
```suggestion
    /// Memory is not initialized. If uninitialized memory is undesirable, set it
```

rustler/src/types/binary.rs
```diff
@@ -1,57 +1,147 @@
...
+//! Safe wrappers around Erlang binaries.
+//!
+//! Rustler provides two binary types: [`Binary`] and [`OwnedBinary`].  Both
+//! represent a contiguous region `u8`s, and they both use the Erlang allocator. The
+//! primary difference between the two is their ownership semantics.
+//!
+//! The _owned_ in `OwnedBinary` refers to the fact that it owns the binary it
+//! wraps. The _owner_ of an `OwnedBinary` is free to modify its contents. Ownership
+//! lasts until it is dropped or consumed by converting it into a regular
+//! `Binary`. An `OwnedBinary` can not be copied or cloned and is thus always moved.
```
evnu, line 10:
```suggestion
//! `Binary`. An `OwnedBinary` cannot be copied or cloned and is thus always moved.
```

evnu (approved):
Very nice, thanks! I only found two minor nits.

PR #336, opened 2020-11-03
Fix 'multiple build target' warning

evnu:
This fixes the following warning:

```
warning: <path>/rustler_tests/native/rustler_test/Cargo.toml: file found to be present in multiple build targets: <path>/rustler_tests/native/rustler_test/src/main.rs
```

As far as I understand, `main.rs` in `rustler_test` does not serve a purpose any longer. @scrogson Feel free to close if the file is actually needed.


evnu:
I failed to run the unit tests; the tests try to build binaries from `main.rs`.

PR #339, opened 2020-11-11
Force panic unwind implementation for users of rustler

filmor:
Since safety of this library relies on `std::panic::catch_unwind`, we need to configure `rustler` to force `panic = "unwind"` in its `Cargo.toml` according to https://doc.rust-lang.org/edition-guide/rust-2018/error-handling-and-panics/controlling-panics-with-std-panic.html.

### DIFF
--- a/rustler/Cargo.toml
+++ b/rustler/Cargo.toml
@@ -17,6 +17,12 @@ lazy_static = "1.4"
 rustler_codegen = { path = "../rustler_codegen", version = "0.22.0-rc.0", optional = true}
 rustler_sys = { path = "../rustler_sys", version = "~2.1" }
 
+[profile.dev]
+panic = "unwind"
+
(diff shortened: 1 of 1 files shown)


hansihe:
https://doc.rust-lang.org/cargo/reference/profiles.html

> Cargo only looks at the profile settings in the Cargo.toml manifest at the root of the workspace. Profile settings defined in dependencies will be ignored.

I believe we need to put this in the `Cargo.toml` for the NIFs themselves, maybe update the generator instead?

filmor:
I'll test this quickly, the docs I referenced claim that it should fail to compile if one of our users sets `panic = "abort"`.

filmor:
Hmm, yes, indeed. Then maybe we'll just need to document this prominently somewhere and/or check for it during compilation.

PR #340, opened 2020-11-20
Update rust linker flags to apply to macOS

jtdowney:
The previous target works on Intel based Macs but was causing an issue on the new Apple Silicon based Macs.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,4 +1,4 @@
-[target.x86_64-apple-darwin]
+[target.'cfg(target_os = "macos")']
 rustflags = [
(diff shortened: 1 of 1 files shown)


evnu:
Looks good to me, thanks. We don't have macos in our github actions, I assume you tested this on one of your machines?

scrogson (approved):

jtdowney:
> Looks good to me, thanks. We don't have macos in our github actions, I assume you tested this on one of your machines?

Yes, I tested it on both an Intel Macbook Pro and an Apple Silicon Macbook Pro.

PR #343, opened 2020-11-25
ci: Run only on stable

evnu:
Testing against stable and nightly is useful to find upcoming problems early. But this only works properly if testing happens fairly often in the test environment. As we do not have that much activity on this repository, I propose to only build on stable. This reduces the number of runners needed to execute the tests, and should result in a (slightly) quicker turn-around for pull requests (the windows checks still dwarf all other tests).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -172,8 +172,6 @@ jobs:
 
         rust:
           - stable
-          - beta
-          - nightly
     steps:
       - name: Checkout sources
         uses: actions/checkout@v1


filmor:
We could do a scheduled run against nightly every week or so.

evnu:
> We could do a scheduled run against nightly every week or so.

Yes, that's true. @filmor @scrogson would you prefer that?

Note that I don't know who would receive notifications on failed runs if actions are not associated with a pull request.

scrogson:
> We could do a scheduled run against nightly every week or so.

> would you prefer that?

yeah, that sounds like a good idea.

evnu:
Closing in favor of https://github.com/rusterlium/rustler/pull/344. Lets see if a simple schedule already suffices.

PR #344, opened 2020-11-30
ci: Run jobs on wednesday night

evnu:
Run the CI jobs periodically to ensure that we keep up with nightly.

This is an alternative to https://github.com/rusterlium/rustler/pull/343.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,6 +6,8 @@ on:
     tags: ["*"]
   pull_request:
     branches: ["*"]
+  schedule:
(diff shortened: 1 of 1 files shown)


evnu:
I did not find a simple way to adapt the build matrix based on the event (ie drop nightly and beta unless we are running the periodic schedule). A simple schedule is probably enough anyways.

evnu:
Lets see if this runs next wednesday.

evnu:
Seems like it worked: https://github.com/rusterlium/rustler/actions/runs/424469992

image

Run 279 was scheduled.

PR #346, opened 2021-01-06
Fix clippy lints

evnu:
The periodic CI run from https://github.com/rusterlium/rustler/pull/344 found some clippy lints.

* Prefer collect() over from_iter(): clippy::from-iter-instead-of-collect
* Prefer `From` over `Into`: clippy::from-over-into
* Avoid nested match: clippy::collapsible-match
* Avoid unnecessary wraps: clippy::unnecessary-wraps
* Obey naming convention of `as_`: clippy::wrong-self-convention

### DIFF
--- a/rustler/src/codegen_runtime.rs
+++ b/rustler/src/codegen_runtime.rs
@@ -16,14 +16,14 @@ pub use crate::wrapper::{
 pub use rustler_sys::{TWinDynNifCallbacks, WIN_DYN_NIF_CALLBACKS};
 
 pub unsafe trait NifReturnable {
-    unsafe fn as_returned(self, env: Env) -> NifReturned;
+    unsafe fn into_returned(self, env: Env) -> NifReturned;
 }
 
 unsafe impl<T> NifReturnable for T
 where
     T: crate::Encoder,
 {
-    unsafe fn as_returned(self, env: Env) -> NifReturned {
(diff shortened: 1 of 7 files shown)


rustler/src/codegen_runtime.rs
```diff
@@ -16,14 +16,14 @@ pub use crate::wrapper::{
 pub use rustler_sys::{TWinDynNifCallbacks, WIN_DYN_NIF_CALLBACKS};
 
 pub unsafe trait NifReturnable {
-    unsafe fn as_returned(self, env: Env) -> NifReturned;
```
evnu, line 19: Clippy advises that `as_` should take `&self` or `&mut self` (see [here](https://rust-lang.github.io/rust-clippy/master/#wrong_self_convention)). As we want to take ownership here (or am I mistaken?), the function should be prefixed with `into_` instead.

filmor (approved):

rustler/src/codegen_runtime.rs
filmor: This sounds right to me 👍

`codegen_runtime` is `pub` but hidden, so I wouldn't see this as a breaking change.

hansihe: Agreed, `codegen_runtime` is public because it is used by our macros, but is not a documented or stable API outside of that.

hansihe (approved):

PR #347, opened 2021-02-01
Allow specifying a channel for the system rustc

jflatow:
This is useful if you want to use e.g. `{:system, "+nightly"}`

### DIFF
--- a/rustler_mix/lib/rustler/compiler.ex
+++ b/rustler_mix/lib/rustler/compiler.ex
@@ -48,6 +48,7 @@ defmodule Rustler.Compiler do
   defp make_base_command(:system), do: ["cargo", "rustc"]
   defp make_base_command({:system, channel}), do: ["cargo", channel, "rustc"]
   defp make_base_command({:bin, path}), do: [path, "rustc"]
+
   defp make_base_command({:rustup, version}) do
     if Rustup.version() == :none do
       throw_error(:rustup_not_installed)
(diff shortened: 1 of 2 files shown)


rustler_mix/lib/rustler/compiler.ex
```diff
@@ -43,8 +43,8 @@ defmodule Rustler.Compiler do
   end
 
   defp make_base_command(:system), do: ["cargo", "rustc"]
+  defp make_base_command({:system, channel}), do: ["cargo", channel, "rustc"]
```
evnu, line 46: This should be documented in [rustler.ex](https://github.com/rusterlium/rustler/blob/master/rustler_mix/lib/rustler.ex#L22)

jflatow: Added

evnu:
The format `"+nightly"` might come as a surprise to users: If they use something like `{:system, "nightly"}`, `cargo` will think that the channel is actually an unknown command. We should either document the expected format with `+` or use atoms and translate (`{:system, :nightly}`). What do you think? Documenting the required `+` is probably the simplest thing to do here.

jflatow:
Added some clarification, let me know what you think.

evnu:
I like it, thanks!

evnu:
@jflatow Can you fix the format error?

jflatow:
Should be passing CI now, though I don't see the checks running at all again

evnu:
> Should be passing CI now, though I don't see the checks running at all again

Seems that I need to run that explicitly. CI went through, thank you!

PR #350, opened 2021-03-06
Fix f32 type conversion to be safe

Virviil:
Decoding `f32` is made using `enif_get_double`, which obviously consumes values bigger than `f32:MAX`.
Then, `ret as f32` not limiting the value by it's max, either turn it to be `f32:INFINITY`. This leads to confusions and even nif panics.

### DIFF
--- a/rustler/src/types/primitive.rs
+++ b/rustler/src/types/primitive.rs
@@ -60,7 +60,6 @@ impl<'a> Decoder<'a> for bool {
     }
 }
 
-
 impl Encoder for f32 {
     fn encode<'a>(&self, env: Env<'a>) -> Term<'a> {
         #[allow(clippy::cast_lossless)]
@@ -86,8 +85,7 @@ impl<'a> Decoder<'a> for f32 {
         // Values bigger then f32 are coerced as infinity
         if res.is_finite() {
             Ok(res)
-        }
-        else {
+        } else {
             Err(Error::BadArg)
         }
     }
--- a/rustler/src/types/primitive.rs
+++ b/rustler/src/types/primitive.rs
@@ -42,7 +42,6 @@ impl_number_transcoder!(i8, i32, enif_make_int, enif_get_int);
 impl_number_transcoder!(u8, u32, enif_make_uint, enif_get_uint);
 impl_number_transcoder!(i16, i32, enif_make_int, enif_get_int);
 impl_number_transcoder!(u16, u32, enif_make_uint, enif_get_uint);
-impl_number_transcoder!(f32, f64, enif_make_double, enif_get_double);
 impl_number_transcoder!(usize, u64, enif_make_uint64, enif_get_uint64);
 impl_number_transcoder!(isize, i64, enif_make_int64, enif_get_int64);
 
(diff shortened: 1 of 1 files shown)


rustler/src/types/primitive.rs
```diff
@@ -60,3 +59,34 @@ impl<'a> Decoder<'a> for bool {
         atom::decode_bool(term)
     }
 }
+
+impl Encoder for f32 {
+    fn encode<'a>(&self, env: Env<'a>) -> Term<'a> {
+        #[allow(clippy::cast_lossless)]
```
evnu, line 65: Is this needed because of `*self as f64`? If so, `f64::from(*self)` would probably be preferable.

rustler/src/types/primitive.rs
```diff
@@ -60,3 +59,34 @@ impl<'a> Decoder<'a> for bool {
...
+                env,
+                rustler_sys::enif_make_double(env.as_c_arg(), *self as f64),
+            )
+        }
+    }
+}
+impl<'a> Decoder<'a> for f32 {
+    fn decode(term: Term) -> NifResult<f32> {
+        #![allow(unused_unsafe)]
+        let mut res: f64 = Default::default();
```
evnu, line 77: We can get rid of the `unsafe` block and call to `enif_get_double` if we decode to `f64` directly:

```rust
let res: f64 = term.decode()?;
```

rustler/src/types/primitive.rs
```diff
@@ -60,3 +59,34 @@ impl<'a> Decoder<'a> for bool {
...
+        if unsafe {
+            rustler_sys::enif_get_double(term.get_env().as_c_arg(), term.as_c_arg(), &mut res)
+        } == 0
+        {
+            return Err(Error::BadArg);
+        }
+        let res = res as f32;
```
evnu, line 84: I don't know enough about the behaviour of `as` w.r.t. floating point numbers. I guess this results in a potential precision loss and not only a loss in the range of maximum values? Is this something users would be aware of?

rustler/src/types/primitive.rs
```diff
@@ -60,3 +59,34 @@ impl<'a> Decoder<'a> for bool {
...
+        // Values bigger then f32 are coerced as infinity
```
evnu, line 85:
```suggestion
        // Values bigger than f32 are coerced as infinity
```

evnu (changes requested):
Thanks for your pull request! Sorry for taking so long for the review.

rustler/src/types/primitive.rs
Virviil: It seems that converting from f64 to f32 leads to precision loss, but:

1. The same attitude is ALREADY presented in `rustler` (but generated by macro).
2. Probably, library user knows for himself better WHY he's trying to use `f32` instead of `f64` while erlang provides all the floating point values as `doubles`. The only reason can be "to speed up a math", and for that case it seems to be a reasonable tradeoff.

As an alternative, possible solution will be totally remove `f32` as a type that implements `rustler::{Encoder, Decoder}` traits.

evnu: Yeah, I think you are right, users should be aware of this already.

Virviil:
I'm not sure why `clippy` check fail, it will be perfect if someone can check it out.

evnu (changes requested):
Looks good to me! Can you rebase on master to get rid of the clippy lint? After that we are good to go.

evnu:
Thanks again @Virviil !

PR #356, opened 2021-05-04
Duplicate "add" function in the docs.

likeanocean:
Just a trivial change. The second, duplicated `add` should actually be a `sub`.


evnu:
Thanks!

PR #358, opened 2021-05-13
Update supported NIF API version to 2.16

### LINKED ISSUE #357: Add 2.16 to the list of supported NIF versions
The number was bumped to 2.16 in OTP 24.  Adding that to the list of supported versions in rustler/build.rs appears to let things work?  At least, rustler builds and passes its test suite that way, and my own project built with a rustler with that patch does too.  Is there anything more that needs to be done to check compatibility?  If not, I have a pair of branches (one for backports and one for master) ready that add it to the list and which can be turned into pull requests if you want.

filmor:
Fixes #357.


rustler_sys/src/rustler_sys_api.rs
```diff
@@ -138,13 +138,19 @@ pub type ErlNifResourceDown = unsafe extern "C" fn(
...
+    unsafe extern "C" fn(env: *mut ErlNifEnv, obj: *mut c_void, call_data: *const c_void) -> ();
+
 /// See [ErlNifResourceTypeInit](http://www.erlang.org/doc/man/erl_nif.html#ErlNifResourceTypeInit) in the Erlang docs.
 #[derive(Debug, Copy, Clone)]
 #[repr(C)]
 pub struct ErlNifResourceTypeInit {
     dtor: *const ErlNifResourceDtor,
     stop: *const ErlNifResourceStop, // at ERL_NIF_SELECT_STOP event
     down: *const ErlNifResourceDown, // enif_monitor_process
+    members: c_int,
```
evnu, line 152: From the [documentation](https://erlang.org/doc/man/erl_nif.html#enif_init_resource_type):

> The struct also contains the field `members` that must be set to the number of initialized callbacks counted from the top of the struct. For example, to initialize all callbacks including dyncall, members should be set to 4. All callbacks are optional and may be set to NULL. 

I am unsure where we initialize `members`. Is this set to zero in `initmacro.rs`? (I also wonder why `members` is actually needed, if the function pointers can be `NULL`, but well..)

filmor: We don't use `ErlNifResourceTypeInit` in `rustler` as far as I know, so this is only relevant for users that use `rustler_sys` directly.

evnu: Ah, that makes sense! Thanks for the clarification.

deepankar-j:
Is there anything holding back this change from being merged?

filmor:
Nope :)

deepankar-j:
Thanks @filmor!  Is there a planned release date for v0.22?  This fix will hold my project back from moving to OTP 24.

PR #361, opened 2021-06-01
Add workaround for missing lib

### LINKED ISSUE #326: Rustler Compiled Library is Missing
Hi, thanks for starting this great open source project!

## Brief

I am using rustler as a nested dependency, and since `v0.22.0rc`, which I am using via pulling `master` branch, I think the new compilation approach doesn't copy the files correctly when creating a release. It works when I try to run the same release command twice or more, though.

### Code

```elixir
defmodule DamaLogic.NifBridge do
  use Rustler, otp_app: :dama_logic, crate: "damalogic_damaprocess"

  [...]
end
```

Running `mix release` for the first time in a clean docker environment.

#### During compilation

```
226	
227	03:31:43.683 [warn]  The on_load function for module Elixir.DamaLogic.NifBridge returned:
228	{:error, {:load_failed, 'Failed to load NIF library: \'/work/_build/prod/lib/dama_logic/priv/native/libdamalogic_damaprocess.so: cannot open shared object file: No such file or directory\''}}
229	
230	Generated dama_logic app
231	
```

#### After compilation

Running `find . -iname '*.so'`

```
./deps/dama_logic/priv/native/libdamalogic_damaprocess.so

./_build/prod/rel/dama_server/lib/runtime_tools-1.14/priv/lib/trace_ip_drv.so
./_build/prod/rel/dama_server/lib/runtime_tools-1.14/priv/lib/dyntrace.so
./_build/prod/rel/dama_server/lib/runtime_tools-1.14/priv/lib/trace_file_drv.so
./_build/prod/rel/dama_server/lib/crypto-4.6.5/priv/lib/crypto.so
./_build/prod/rel/dama_server/lib/crypto-4.6.5/priv/lib/crypto_callback.so
./_build/prod/rel/dama_server/lib/crypto-4.6.5/priv/lib/otp_test_engine.so

./_build/prod/lib/dama_logic/native/damalogic_damaprocess/release/deps/libdamalogic_damaprocess.so
./_build/prod/lib/dama_logic/native/damalogic_damaprocess/release/deps/librustler_codegen-1426f88bc646ec0e.so
./_build/prod/lib/dama_logic/native/damalogic_damaprocess/release/libdamalogic_damaprocess.so
```

(Newlines added by me for clarity)

## Expectation

I would expect the output of the `*.so` to be within the `_build` directory instead of the `deps` directory given that `:build_embedded` is set to `true`. If the `priv` copying is happening before compilation, I guess this would cause the issue of breaking only during the first attempt of generating a release?

I will attempt to make a small repro repo once I get around to it.

Thanks,

evnu:
This fixes f01719c: We accidentally removed a workaround for mix, which copied the libraries to the target directory in `_build`. d053522f originally introduced a workaround, copying `symlink_or_copy`. Here, we call into a public function of `Mix.Project` instead (which might be more future proof).

Fix #326

### DIFF
--- a/rustler_mix/lib/rustler/compiler.ex
+++ b/rustler_mix/lib/rustler/compiler.ex
@@ -37,7 +37,7 @@ defmodule Rustler.Compiler do
       end
 
       handle_artifacts(crate_full_path, config)
-      workaround_mix_priv_dir(config)
+      workaround_mix_priv_dir(config.__project_config__)
     end
 
     config
@@ -131,12 +131,12 @@ defmodule Rustler.Compiler do
 
   # Workaround for a mix problem. We should REALLY get this fixed properly. Originally introduced
   # with d053522fe8b08bdacacb64592b22536e23ff3853.
-  defp workaround_mix_priv_dir(config) do
+  defp workaround_mix_priv_dir(project_config) do
     _ =
       symlink_or_copy(
-        config,
+        project_config,
         Path.expand("priv"),
-        Path.join(Mix.Project.app_path(config), "priv")
+        Path.join(Mix.Project.app_path(project_config), "priv")
       )
   end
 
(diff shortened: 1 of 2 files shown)


Mazyod (approved):
build passes on my repro repo 👍

filmor (approved):

PR #376, opened 2021-08-21
[wip] add support for monitors

dlesl:
This is still at a PoC stage but I'm opening this PR to get some feedback :). There's at least one thing that I think definitely needs some work, the naming of `ResourceArcMonitor`, `Monitor` and `MonitorResource`.


rustler/src/resource.rs
```diff
@@ -227,6 +264,50 @@ where
...
+        }
+    }
+
+    fn demonitor(&self, caller_env: Option<&Env>, mon: &Monitor) -> bool {
+        let env = maybe_env(caller_env);
+        unsafe { rustler_sys::enif_demonitor_process(env, self.raw, &mon.inner) == 0 }
+    }
+}
+
+fn maybe_env(caller_env: Option<&Env>) -> NIF_ENV {
```
dlesl, line 292: I'm not sure if this is necessary. The docs state `Argument caller_env is the environment of the calling thread (process bound or callback environment) or NULL if calling from a custom thread not spawned by ERTS.`. However looking at the source code, `env` isn't even used. But in older versions of OTP (and maybe in the future again?) it was used, so it might be safer to keep this logic (which is currently just a copy-paste from OwnedEnv)

rustler/src/resource.rs
```diff
@@ -65,6 +82,24 @@ extern "C" fn resource_destructor<T>(_env: NIF_ENV, handle: MUTABLE_NIF_RESOURCE
...
+) {
+    unsafe {
+        let pid = LocalPid::new((&*pid).clone());
+        let mon = Monitor { inner: (&*mon).clone() };
+        crate::wrapper::resource::keep_resource(handle);
+        let resource = ResourceArc {
+            inner: align_alloced_mem_for_struct::<T>(handle) as *mut T,
+            raw: handle
+        };
+        T::down(&resource, pid, mon);
```
dlesl, line 99: Probably this shouldn't be a reference

rustler_sys/src/rustler_sys_api.rs
```diff
@@ -146,11 +146,11 @@ pub type ErlNifResourceDynCall =
 #[derive(Debug, Copy, Clone)]
 #[repr(C)]
 pub struct ErlNifResourceTypeInit {
-    dtor: *const ErlNifResourceDtor,
-    stop: *const ErlNifResourceStop, // at ERL_NIF_SELECT_STOP event
-    down: *const ErlNifResourceDown, // enif_monitor_process
-    members: c_int,
-    dyncall: *const ErlNifResourceDynCall,
+    pub dtor: Option<ErlNifResourceDtor>,
```
evnu, line 149: Is `pub(crate)` enough?

rustler/src/resource.rs
evnu: If the documentation says so, I think we should stick with the documented behaviour. But maybe we could turn this into a helper function instead of duplicating code? E.g. `Env::called_from_thread(env)`?

evnu (commented):
Thanks for this PR! I like the direction and have some preliminary comments and questions.

rustler/src/resource.rs
```diff
@@ -65,6 +82,24 @@ extern "C" fn resource_destructor<T>(_env: NIF_ENV, handle: MUTABLE_NIF_RESOURCE
...
+pub extern "C" fn resource_down<T: MonitorResource>(
+    _env: NIF_ENV,
+    handle: MUTABLE_NIF_RESOURCE_HANDLE,
+    pid: *const ErlNifPid,
+    mon: *const ErlNifMonitor
+) {
+    unsafe {
+        let pid = LocalPid::new((&*pid).clone());
+        let mon = Monitor { inner: (&*mon).clone() };
+        crate::wrapper::resource::keep_resource(handle);
```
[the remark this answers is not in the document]
dlesl: `enif_release_resource` is called by `ResourceArc`'s `Drop` impl when `resource` is dropped

rustler/src/resource.rs
```diff
@@ -65,6 +82,24 @@ extern "C" fn resource_destructor<T>(_env: NIF_ENV, handle: MUTABLE_NIF_RESOURCE
...
+        let resource = ResourceArc {
+            inner: align_alloced_mem_for_struct::<T>(handle) as *mut T,
```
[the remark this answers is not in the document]
dlesl: According to the docs the BEAM guarantees that this pointer will be valid, so I implemented this the same way as `resource_destructor` (which has the same guarantee)

dlesl:
Thanks for the review! I'll try to implement your suggestions soon(ish) :)

filmor:
Thank you for your contribution. I took most of your tests, but had to implement things a bit differently to make it work with the refactored resource implementation.

PR #378, opened 2021-09-04
codegen: change RaiseAtom to RaiseTerm

### LINKED ISSUE #375: Unit enum decoding failure returns an atom instead of raising an exception
Hi and thanks for this awesome library! I just wanted to say that I found it a bit surprising that the code generated by `rustler_codegen` returns `Err(rustler::Error::Atom("invalid_variant"))` for unit enums, but in other cases (for example a missing key in a map) it returns `Err(rustler::Error::RaiseTerm(...))`. Would you be open to changing this behaviour?

dlesl:
fixes #375


filmor (approved):

filmor:
Could you please add a Changelog entry for this change?

hansihe:
This is also a breaking change, which is something we need to consider.

dlesl:
> This is also a breaking change, which is something we need to consider.

I mentioned this in the Changelog, do you think that's enough?

PR #379, opened 2021-09-04
add `NewBinary` (`enif_make_new_binary`)

dlesl:
It seems like this can improve performance a bit if you're creating a lot of small binaries

### DIFF
--- a/rustler/src/types/binary.rs
+++ b/rustler/src/types/binary.rs
@@ -1,8 +1,9 @@
 //! Safe wrappers around Erlang binaries.
 //!
-//! Rustler provides two binary types: [`Binary`] and [`OwnedBinary`].  Both
-//! represent a contiguous region `u8`s, and they both use the Erlang allocator. The
-//! primary difference between the two is their ownership semantics.
+//! Rustler provides three binary types: [`Binary`], [`NewBinary`] and
+//! [`OwnedBinary`]. All represent a contiguous region `u8`s, and they all use
+//! the Erlang allocator. The primary difference between them is their ownership
+//! semantics.
 //!
 //! The _owned_ in `OwnedBinary` refers to the fact that it owns the binary it
 //! wraps. The _owner_ of an `OwnedBinary` is free to modify its contents. Ownership
@@ -13,6 +14,12 @@
 //! cheap to copy: all copies of a `Binary` point to the original `Binary`'s
 //! data. Additionally, a `Binary`'s lifetime is tied to that of the NIF's [`Env`],
 //! preventing outstanding references to the data after a NIF returns.
+//!
+//! `NewBinary` is a way of creating a `Binary` without going via `OwnedBinary`.
+//! This can improve performance, since `NewBinary`s can be allocated on the
+//! heap if they are small. Unlike `OwnedBinary`, `NewBinary`s lifetime is tied
+//! to that of the NIF's [`Env`]. `NewBinary` must be converted to a `Binary`
+//! or directly to a `Term` before it can be passed to Erlang.
 //!
 //! # Examples
 //!
@@ -79,7 +86,7 @@
 //! [`OwnedBinary`]: struct.OwnedBinary.html
 
 use crate::{
-    wrapper::binary::{alloc, realloc, ErlNifBinary},
+    wrapper::binary::{alloc, new_binary, realloc, ErlNifBinary},
     Decoder, Encoder, Env, Error, NifResult, Term,
 };
 use std::{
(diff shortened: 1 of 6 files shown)


evnu:
> It seems like this can improve performance a bit if you're creating a lot of small binaries

Out of curiosity: Have you done benchmarks to compare the performance?

hansihe:
Also, what about calling it something like `EnvBinary` or `ProcessBinary` in order to make things a bit clearer in the name?

dlesl:
> Out of curiosity: Have you done benchmarks to compare the performance?

I had a go at it like this (erlang just because I'm not so comfortable with elixir):

``` erlang
-module(bench).

-export([run/1]).
-define(TIME, 10000).

run(NumWorkers) ->
  [bench(fun 'Elixir.RustlerTest':F/0, NumWorkers)
   || F <- [binary_new, owned_binary_new, new_binary_new]].

bench(Fun, NumWorkers) ->
  io:format(">> testing ~p, ~B workers~n", [Fun, NumWorkers]),
  Self = self(),
  Start = os:timestamp(),
  Workers = [spawn(fun() -> worker(Self, Fun, 0) end) || _ <- lists:seq(1, NumWorkers)],
  timer:sleep(?TIME),
  [W ! stop || W <- Workers],
  Results = [receive Res -> Res end || _ <- lists:seq(1, NumWorkers)],
  DiffMs = timer:now_diff(os:timestamp(), Start) div 1000,
  Calls = lists:sum(Results),
  io:format("~B calls in ~B ms, ~f calls/s~n", [Calls, DiffMs, Calls / (DiffMs / 1000.0)]).

worker(Pid, Fun, Calls) ->
  receive
    stop -> Pid ! Calls
  after 0 ->
      <<1, 2, 3, 4>> = Fun(),
      worker(Pid, Fun, Calls + 1)
  end.
```
Then, in `rustler_test`,
```
$ MIX_ENV=prod iex -S mix
...
iex(1)> c("bench.erl")
[:bench]
iex(2)> :bench.run(1); :bench.run(10)
>> testing fun 'Elixir.RustlerTest':binary_new/0, 1 workers
38011410 calls in 10000 ms, 3801141.000000 calls/s
>> testing fun 'Elixir.RustlerTest':owned_binary_new/0, 1 workers
40077911 calls in 10000 ms, 4007791.100000 calls/s
>> testing fun 'Elixir.RustlerTest':new_binary_new/0, 1 workers
68728757 calls in 10000 ms, 6872875.700000 calls/s
>> testing fun 'Elixir.RustlerTest':binary_new/0, 10 workers
75336109 calls in 10002 ms, 7532104.479104 calls/s
>> testing fun 'Elixir.RustlerTest':owned_binary_new/0, 10 workers
76902124 calls in 10001 ms, 7689443.455654 calls/s
>> testing fun 'Elixir.RustlerTest':new_binary_new/0, 10 workers
140989521 calls in 10001 ms, 14097542.345765 calls/s
[:ok, :ok, :ok]
```

dlesl:
> Also, what about calling it something like `EnvBinary` or `ProcessBinary` in order to make things a bit clearer in the name?

Yea it's a bit of a tricky one to name. I went with `NewBinary` because that's similar to upstream and it captures the fact that you can only get one of these by creating it. With `EnvBinary` would you expect to be able to get one as a reference to an existing binary? And I don't think `ProcessBinary` is quite right either, because you can create one of these on a process-independent `OwnedEnv` (useful for example if you want to send a small binary in a message).

evnu:
From the [documentation](https://www.erlang.org/doc/man/erl_nif.html#enif_make_monitor_term):

> it cannot be reallocated.

What does that mean for this type? Are special precautions necessary?

dlesl:
> From the [documentation](https://www.erlang.org/doc/man/erl_nif.html#enif_make_monitor_term):
> 
> > it cannot be reallocated.
> 
> What does that mean for this type? Are special precautions necessary?

My understanding is that it means you can't change the size of the allocation later (for example using `enif_realloc_binary`. Since we don't expose this functionality I don't think we need to take any precautions.

evnu:
I think this PR mostly hangs on the type's name. Two more ideas:

* `MutBinary` - `Binary` is an immutable smart-pointer, so `MutBinary` is a mutable smart-pointer
* `RawBinary` - As it exposes a binary's raw inner data

@dlesl Would you mind rebasing on master?

dlesl:
Rebased. I agree that the naming is the most difficult part. Thanks for the additional suggestions. I like `MutBinary`, but I'm not sure that it sufficiently captures the difference to `OwnedBinary`, which is also mutable?

evnu:
> I like MutBinary, but I'm not sure that it sufficiently captures the difference to OwnedBinary, which is also mutable?

You are right, that would be misleading as well. How does `MutEnvBinary` sound? It is a mutable binary which is bound to an environment.

dlesl:
Hmm, that could work, but I don't know if it captures the advantages (performance). I still have a preference for going with the upstream naming -- it makes it easier to find it when you know you're looking for it (after reading the erl_nif docs for example).

evnu (approved):

evnu:
Lets merge this, we can rename later if we find a more suitable name.

PR #383, opened 2021-09-11
Add NifException macro

thomas9911:
I was trying to map and Rust struct to a Elixir exception. After some trial and error I found out that elixir adds an __exception__ field to a struct with the value true. When you add this in Rust it maps correctly to the exception module. I thought it would be handy to have this feature in Rustler.

Ps: I dont have much experience with proc_macro, so if there are things that can be improved I would like to hear it.


rustler_codegen/src/ex_struct.rs
```diff
@@ -18,11 +18,22 @@ pub fn transcoder_decorator(ast: &syn::DeriveInput) -> TokenStream {
     // Unwrap is ok here, as we already determined that struct_fields is not None
     let field_atoms = ctx.field_atoms().unwrap();
 
-    let atom_defs = quote! {
-        rustler::atoms! {
-            atom_struct = "__struct__",
-            atom_module = #elixir_module,
```
evnu, line 24: Maybe the exception atom could be quoted as well and inserted here. Then we don't need to duplicate `rustler::atoms!`. Something along these lines (untested!):

```rust
exception_quote =
if add_exception {
  quote!{atom_exception = "__exception__"}
} else {
  quote!{}
}
//...

rustler::atoms! {
  //...
  #exception_atom,
}
```

evnu: Ah, that's actually quite similar to the encoding case (`exception_field`).

rustler_codegen/src/ex_struct.rs
```diff
@@ -5,7 +5,7 @@ use syn::{self, spanned::Spanned, Field, Ident};
 use super::context::Context;
 use super::RustlerAttr;
 
-pub fn transcoder_decorator(ast: &syn::DeriveInput) -> TokenStream {
+pub fn transcoder_decorator(ast: &syn::DeriveInput, add_exception: bool) -> TokenStream {
```
evnu, line 8: I am not a huge fan of a boolean `add_exception` flag for this, but aside from copying the module as a whole, I can't think of a cleaner way. @rusterlium/core maybe someone else has an idea to avoid this flag?

evnu:
@thomas9911 Thanks! Only formatting left now :)

thomas9911:
classic, done!

evnu (approved):

PR #270, opened 2019-10-27
Revamp rustler_mix documentation

evnu:
I restructured the documentation for rustler_mix in an attempt to make it more accessible on hex.pm. For this, I merged the basic guide into the `README.md`, made the `README.md` as the main file for `ex_doc`, replaced `def` with `defp` for internal functions, and removed internal modules from the output. Let me know what you think!

PR #384, opened 2021-09-13
Point to HexDocs base URL for Rustler in lib.rs

brainsnail:
Hey folks! 

Was checking this out and I noticed that the URL in `lib.rs` is pointing to a dead link. I _think_ it's related to #270 as it looks like this is when the `mix` documentation was updated. I compared the docs and it looks like the base Rustler module docs cover pretty much everything there. Happy to update this PR if I'm mistaken though.

Thanks for making the project 😄


evnu:
Thanks!

PR #387, opened 2021-10-06
Define supported versions of OTP and Elixir

evnu:
This pull request defines which versions of OTP and Elixir Rustler aims to support. This definition is then applied to remove fallback code for `make_map_from_arrays()`.


hansihe (approved):
Nice, LGTM!

scrogson (approved):

PR #389, opened 2021-10-06
Don't call cargo when skip_compilation? is true

karolsluszniak:
This fixes a regression from #386 which made rustler_mix call `cargo metadata` even when skip_compilation? is set to true,
breaking setups in which the rust library is built separately, e.g. in separate docker stage with proper rust versioning and in which the elixir compilation stage doesn't have cargo.

This fix assumes that the external_resources attribute is irrelevant when skip_compilation? is set to true.

### DIFF
--- a/rustler_mix/lib/rustler/compiler/config.ex
+++ b/rustler_mix/lib/rustler/compiler/config.ex
@@ -61,9 +61,13 @@ defmodule Rustler.Compiler.Config do
     crate = Keyword.fetch!(opts, :crate)
 
     resources =
-      opts
-      |> Keyword.get(:path)
-      |> external_resources(crate)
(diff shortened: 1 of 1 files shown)


evnu:
Thanks for the pull request! I wonder if we should instead skip calling into `Rustler.Compiler.compile_crate/2` altogether if `skip_compilation?` is set. @scrogson what do you think?

evnu:
> > I wonder if we should instead skip calling into Rustler.Compiler.compile_crate/2 altogether if skip_compilation? is set.
> 
> @evnu makes sense to me.

I think we should go ahead with this PR as is and release a patch version, and refactor later.

evnu:
@karolsluszniak I released v0.22.2 with this fix. Thanks again!

PR #393, opened 2021-10-13
ci: Test against v1.12

evnu:
CI should also test Elixir v1.12.


evnu:
This PR should work after #399.

Issue #388, opened 2021-10-06
Port gen_api.erl to Rust so it can be run without escript installed


hansihe:
As far as I can see, this only accesses some `system_info`, most of which is inferable from the target platform.

Other things, like dirty NIF support might not even be that relevant anymore.

PR #397, opened 2021-10-22
Rewrite build of NIF APIs in Rust

philss:
This PR changes the way the NIF APIs are generated to be included in the `rustler_sys_api.rs` file by replacing the script that was written in Erlang to one written in Rust.

We need this change in order to simplify the cross-compilation of Rustler crates for different NIF version targets.

Before this change we were using an Erlang script to get which **NIF version** to use in the build process, and this required the presence of Erlang installed in the build system.

Now the dependency of Erlang is ~~removed~~ optional and ~~a new requirement was added: in order to build  a crate with target to `release`~~ **the environment variable `RUSTLER_NIF_VERSION` ~~must~~ should be present**. **EDIT: if `RUSTLER_NIF_VERSION` is not present, then we fetch that from Erlang or we use the latest available version. In the process of build creates for an Elixir project, the compiler will inject that env var so nothing changes for users.** 

~~This new requirement is not a problem if one is building a dependency from the Elixir side using the Rustler compiler, because we set that value when building the lib.~~ (**EDIT: is not a requirement anymore.**)

## Acknowledgements

Most of the `rustler_sys/build.rs` was work of @hansihe. Thank you for that! :D

I'm a newbie in Rust, but my friend @lsunsi helped me a lot with this PR. Thanks, migo! <2

~~PS: this PR includes some changes from https://github.com/rusterlium/rustler/pull/395 . After merge that one I should rebase this.~~ **done**.
PS2: most of the changes are in the `rustler_sys/build.rs` file. 
~~PS3: open as draft because I want to try adding some tests to that file.~~ **EDIT: I couldn't test the `build.rs` directly, but I think that relying on the compilation is good enough for now.**.

**EDIT: this closes https://github.com/rusterlium/rustler/issues/388**


filmor:
Thank you very much :)

PR #400, opened 2021-11-02
Implement Hash and [Partial]Eq for binary types

JayKickliter:
I have a concrete use-case for this PR: using an `OwnedBinary` as the [key] in caches/maps. I'm currently forced to wrap the binary in a clumsy wrapper struct.

For completeness sake, this PR implements Hash, PartialEq, and Eq for Binary in addition to implementing PartialEq between Binary ⟷ OwnedBinary.

[key]: https://github.com/helium/e2qc/pull/1/files#diff-28b0ccc836e6e883168b88ebb6f39508969ece5516e5f81458abd33592b12430R195-R221

### DIFF
--- a/rustler/src/types/binary.rs
+++ b/rustler/src/types/binary.rs
@@ -84,6 +84,7 @@ use crate::{
 };
 use std::{
     borrow::{Borrow, BorrowMut},
+    hash::{Hash, Hasher},
     io::Write,
     mem::MaybeUninit,
     ops::{Deref, DerefMut},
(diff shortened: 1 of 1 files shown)


JayKickliter:
>  requested a review from yesterday

That's weird. From who? A GitHub bug?

evnu:
> That's weird. From who? A GitHub bug?

image

:arrow_up: That's what it should look like :)

evnu:
LGTM, I wonder if we should have a unit test for this?

hansihe:
I think this is good as-is, we don't necessarily need a unit test for something like this.

PR #401, opened 2021-11-03
ci: Add macos runner

evnu:
Add a CI runner for macOS. On macos, we need to run `rustc` separately for each target in the `Cargo.toml`. The reason is that we currently pass extra linker arguments to `rustc`. This is not possible if multiple targets need to be build.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -181,3 +181,35 @@ jobs:
         run: |
           mix deps.get
           mix test
+
+  test_macos:
+    name: macos-11 test
+    runs-on: macos-11
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v1
+
+      - name: Install Erlang/Elixir
+        run: |
+          brew install elixir
+          mix local.hex --force
+
+      - name: Install Rust ${{matrix.rust}} toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          override: true
+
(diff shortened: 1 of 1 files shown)


evnu:
Compilation on macOS currently fails for the tests, as we pass extra arguments:

```
error: extra arguments to `rustc` can only be passed to one target, consider filtering
the package by passing, e.g., `--lib` or `--bin NAME` to specify a single target

== Compilation error in file lib/rustler_test.ex ==
** (RuntimeError) Rust NIF compile error (rustc exit code 101)
    (rustler 0.22.2) lib/rustler/compiler.ex:36: Rustler.Compiler.compile_crate/2
    lib/rustler_test.ex:6: (module)
    (stdlib 3.16.1) erl_eval.erl:685: :erl_eval.do_apply/6
```

For this to work, we need to call `cargo build`/`cargo rustc` separately for each target.

evnu:
> Compilation on macOS currently fails for the tests, as we pass extra arguments:

I work around that with 65d4171. For macOS, each target (bin or lib) is compiled separately. This feels kind-of hacky.

evnu:
Maybe someone from @rusterlium/core can give some early feedback here.

evnu:
I rebased on #439 to check if the runner works on that branch.

evnu:
As this no longer requires changes to `Rustler.Compiler` and simply adds a runner, I will merge this.

PR #404, opened 2021-11-16
Add FreeBSD support

madninja:
The target OS selection is not actually being used but is stopping rustler from being used on another very unixy system

### DIFF
--- a/rustler_sys/build.rs
+++ b/rustler_sys/build.rs
@@ -23,16 +23,15 @@ trait ApiBuilder {
 }
 
 #[derive(PartialEq, Clone, Copy)]
-pub enum OS {
+pub enum OsFamily {
+    Unix,
     Win,
-    Darwin,
-    Linux,
 }
 
 pub struct GenerateOptions {
     pub ulong_size: usize,
     pub nif_version: (u32, u32),
-    pub target_os: OS,
+    pub target_family: OsFamily,
 }
 
 fn write_ret(out: &mut String, ret: &str) {
@@ -207,7 +206,7 @@ fn generate(opts: &GenerateOptions) -> String {
     .unwrap();
 
     // Basic
-    if opts.target_os == OS::Win {
+    if opts.target_family == OsFamily::Win {
         writeln!(out, "#[allow(dead_code)]").unwrap();
         writeln!(out, "#[derive(Copy, Clone)]").unwrap();
         writeln!(out, "pub struct TWinDynNifCallbacks {{").unwrap();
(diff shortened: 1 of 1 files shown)


evnu:
We may also need to consider this for other BSDs (see https://doc.rust-lang.org/reference/conditional-compilation.html#target_os).

madninja:
> We may also need to consider this for other BSDs (see https://doc.rust-lang.org/reference/conditional-compilation.html#target_os).

It's odd since as far as I can tell the only thing the code uses this for right now is to decide whether it's a unix system or not, which would've been serviced by `target_family`? 

Assuming you already thought through this I can definitely add the other values

filmor:
Yes, this code can be simplified. The only difference is that Windows has a weird construction of "callbacks" instead of just linking the symbols, and that's all there is in the original header. So it's (from my PoV) safe to handle Windows explicitly, and handle all other cases in the same (unix) way.

madninja:
> Yes, this code can be simplified. The only difference is that Windows has a weird construction of "callbacks" instead of just linking the symbols, and that's all there is in the original header. So it's (from my PoV) safe to handle Windows explicitly, and handle all other cases in the same (unix) way.

I'd agree given the seen use in the code. This may be better done as an `OsFamily` instead of an `OS`. I'm not the designer of this though just trying to get back a working rustler on all previously working unix platforms that are now broken

filmor:
> I'd agree given the seen use in the code. This may be better done as an `OsFamily` instead of an `OS`. I'm not the designer of this though just trying to get back a working rustler on all previously working unix platforms that are now broken

Just go ahead with that idea. The separate case for "Darwin", even though it's completely identical to the Linux case, is confusing, replacing the `OS` by `OsFamily::Windows` and `OsFamily::Unix` with wildcard handling of the target os is fine by me and should not break anything (as said, the Erlang header that we are mimicking here does the same).

madninja:
> > I'd agree given the seen use in the code. This may be better done as an `OsFamily` instead of an `OS`. I'm not the designer of this though just trying to get back a working rustler on all previously working unix platforms that are now broken
> 
> Just go ahead with that idea. The separate case for "Darwin", even though it's completely identical to the Linux case, is confusing, replacing the `OS` by `OsFamily::Windows` and `OsFamily::Unix` with wildcard handling of the target os is fine by me and should not break anything (as said, the Erlang header that we are mimicking here does the same).

Ok, let's try that.. pushed the family base proposal

filmor (approved):

rustler_sys/build.rs
```diff
@@ -933,18 +932,17 @@ fn main() {
...
+        env::var("CARGO_CFG_TARGET_FAMILY").unwrap_or_else(|_| env::consts::FAMILY.to_string());
 
-    let target_os = match &target_os_or_current as &str {
-        "windows" => OS::Win,
-        "linux" => OS::Linux,
-        "macos" => OS::Darwin,
+    let target_family = match &target_family_or_current as &str {
+        "windows" => OsFamily::Win,
+        "unix" => OsFamily::Unix,
         other => panic!("Unsupported Operational System: {}", other),
```
philss, line 941: Just a nit: would be nice to change the panic message:

```suggestion
        other => panic!("Unsupported Operational System Family: {}", other),
```

philss (approved):
Only a comment and then :shipit:

madninja:
> Only a comment and then :shipit:

Done.. I'm not sure what the release cycle is, but could we get this merged/tagged relatively soon so we can stop using an old rustler commit?

evnu:
@madninja the current rustler release v0.22.2 does not contain https://github.com/rusterlium/rustler/pull/397 yet, so it should already work for you I think.

madninja:
> @madninja the current rustler release v0.22.2 does not contain #397 yet, so it should already work for you I think.

It does.. v0.22.2 is also quite a bit older than what's on master today?

madninja:
I take it back, apologies.. Oct 6 on the tags is newer than the actual release on the releases page.  

Thanks for the heads up.

PR #408, opened 2021-11-23
Add MapSet -> HashSet decoder/encoder support out of the box

hazardfn:
Adds a new encoder/decoder for the `MapSet.t()/HashSet<K>` type.

Not sure how the rustler team feels about contributions like this that attempt to unwind and rebuild opaque elixir types. There are some obvious drawbacks here:

* The internal representation of a MapSet could change at any time
* Version number used is static in the code
* Depending on how well tested user code is the version used by rustler could result in `==` comparisons failing silently for 2 "identical" mapsets. MapSet.equal?/2 however should always return the right result regardless...

I would be interested in feedback on how best to solve these problems and also your opinions on the value of encoding/decoding opaque types in the elixir stdlib like this...

I mostly did this work just to get hands-on in how encoders/decoders actually work so am happy for this to be closed if it isn't a direction the rustler team wish to head in.

### DIFF
--- a/rustler/src/types/atom.rs
+++ b/rustler/src/types/atom.rs
@@ -308,4 +308,7 @@ atoms! {
 
     /// The `step` atom used by `Elixir.Range` vor Elixir >= v1.12
     step,
+
+    /// The `map` atom as used in `Elixir.MapSet`.
+    map
 }
--- a/rustler/src/types/atom.rs
+++ b/rustler/src/types/atom.rs
@@ -310,5 +310,11 @@ atoms! {
     step,
 
     /// The `map` atom as used in `Elixir.MapSet`.
-    map
+    map,
+
+    /// The `version` atom as used in `Elixir.MapSet`. This atom is flagged for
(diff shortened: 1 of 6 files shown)


josevalim:
Yeah, unfortunately this is relying on private information from MapSet which may change. Especially the assembling of MapSet from inside Rust.

The only way I can think for now is to have the conversion steps in Elixir. For example, a MapSet could become `{:hash_set, MapSet.to_list(set)}` in Elixir (which then becomes HashSet) and on the way back `{:hash_set, keys}` become `MapSet.new(keys)`.

hansihe:
Thanks for the PR!

Unfortunately, as José mentioned, we want to avoid relying on private implementation details like this.

The best way of passing opaque types like this back and fourth would probably indeed be to do some of the conversion work on the Elixir side.

If you really need this for internal use, you could make a newtype for `HashSet` and implement the encoder/decoder for that. You are still relying in `MapSet` implementation details, so in this case you would be responsible for validating/updating every time you update Elixir. It is probably advised to avoid this if you can.

PR #410, opened 2021-11-25
implement hashing for term (rebased)

hazardfn:
As per #189 I have done a simple rebase of this feature. 

It's unclear if more work is required but I'm happy to do it if more needs to be done.


hansihe (approved):
Thanks for the PR, this looks great to me!

Issue #13, opened 2016-04-28
error when reload the module. 


agelin:
Hello.Thanks for the great project.

I found the error when I reload nif module on iex.

```
The on_load function for module Elixir.Test returned {{:badmatch,
  {:error, {:upgrade, 'Upgrade not supported by this NIF library.'}}},
```

Is that has some way to add/edit/export the upgrade function?

Thank you.

hansihe:
Yeah, I did not implement module reloading, as I don't currently see a way to maintain memory when it comes to resource structs across reloads.

I could add the option of allowing upgrades, but it would invalidate memory safety if you modify resource structs and reload the nif library. If you are not using resource structs, it would probably not be a problem.

agelin:
Ok, thanks for your reply!

hansihe:
I don't want to shut down the discussion around this or anything. If there is a need for this it should certainly be bumped up the feature priority list.

agelin:
Last night I think about your anower. On reflection, I think you are right. 

Of cause,“upgrade” may be abused.

Close（/hide）some danger options, to keep the safe mode. It's always effective.

If possible, can there be more a clear full marco of insecurity？like “rustler_export_nifs!!!”，😄. developers can have more choices and know the consequences.

Thanks for the great job.

jorendorff:
Upgrades probably can't be made safe. But we could probably add some kind of `unsafe upgrade: my_upgrade_fn` option to `rustler_export_nifs`. (Requiring the user to type in the `unsafe` keyword is good enough for me.)

bigfish24:
Sorry if this is a basic question, but without support for NIF upgrades, what happens when you perform an upgrade?

davydog187:
I believe the module will fail to load @bigfish24

hansihe:
I believe what I wrote earlier in this thread is wrong.

As long as we always create resource types with the `ERL_NIF_RT_CREATE` flag, I believe we can do upgrades pretty safely.

Issues start to arise when we want to try supporting `ERL_NIF_RT_TAKEOVER`, since that's where we have the inability to prove that whatever is within the resource is identical between the two versions.

Having a basic version of NIF upgrading would certainly be better than having nothing at all

PR #413, opened 2021-12-06
Add dynamic load data

kaaboaye:
Makes it possible to compute `load_data` at the runtime.

Example use case it to provide priv path to the nif.

```ex
defmodule NIF do
  use Rustler, load_data_fun: {Deployment, :nif_data}
end

defmodule Deployment do
  def nif_data do
    :code.priv_dir(:surfer_local) |> IO.iodata_to_binary()
  end
end
```

Also added explicit errors to indicate that user provided invalid config.

image
image


hansihe:
I like the idea here, but I think dynamic `load_data` should be a separate argument, something like `load_data_fun`?

As is, this will break existing `load_data`s that are atom pairs.

kaaboaye:
Sure, I'll update it in the free time and ping you

kaaboaye:
@hansihe refactored and tested on my project

evnu:
@kaaboaye do you think you could add a test for this? In `rustler_test`, any NIF could also load some data with that function. We could then test that this data is available.

kaaboaye:
I've added test making use of dynamic load.

 However, is this output with a couple of panics is expected?

image

evnu:
Thanks for the tests! Yes, the output is expected.

evnu:
> Should I fix it for windows or figure out something different?

I think fixing it for windows is sufficient. We have runners for linux, macos and windows, it suffices to only make them work for now.

rustler_tests/native/dynamic_load/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+    eprintln!("Loading dataset from {:?}.", &asset_path);
+
+    let data = read_to_string(asset_path).unwrap().into_boxed_str();
+    let data = Some(data);
+
+    // Safety: assumes that this function is being called once when
+    // dynamically loading this library.
+    // `load()` is being called exactly once and OTP will not allow any other function call
+    // before this function returns
+    unsafe { DATASET = data };
```
scrogson, line 19: I think we can remove the need for unsafe here if we use `once_cell`?

kaaboaye: Relying on external synchronization mechanisms is unsafe by definition. We need to use unsafe here because rust compiler doesn't know about synchronization guarantees provided by OTP's NIF implementation.

If we were to use any safe rust code it would need to provide its own synchronization which defeats the purpose of relying on OTP's synchronisation or lie to the compiler about synchronization which defeats the purpose of unsafe {} keyword.

Also [std::cell::OnceCell](https://doc.rust-lang.org/std/cell/struct.OnceCell.html) is in nightly so it's a no go anyway.

evnu: > If we were to use any safe rust code it would need to provide its own synchronization which defeats the purpose of relying on OTP's synchronisation

But that would be ok, wouldn't it? As long as we can guarantee that the value is only written once, we are safe I think?

I wonder if we can somehow forbid that `DATASET` is written twice, without using `once_cell`. But we can probably revisit that later? Note that examples from the tests might be reused by users.

evnu: @scrogson I am ok with using unsafe here. `once_cell` being nightly is enough reason for me to stick with `unsafe`. What do you think?

kaaboaye: > I wonder if we can somehow forbid that DATASET is written twice, without using once_cell. But we can probably revisit that later? Note that examples from the tests might be reused by users.

We could put `DATASET` as private member of some module and expose `public(crate)` api for initialization and reading.

Regarding synchronization done inside the rust code we could in theory use [std::sync::RwLock](https://doc.rust-lang.org/stable/std/sync/struct.RwLock.html) whoever it can be poisoned and [LockResult](https://doc.rust-lang.org/stable/std/sync/type.LockResult.html)<[RwLockReadGuard](https://doc.rust-lang.org/stable/std/sync/struct.RwLockReadGuard.html)<'_, T>> also doesn't look encouraging.

If we were to ensure the DATASET is only initialized once we could use [std::sync::atomic:AtomicPtr::compare_exchange_weak](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicPtr.html#method.compare_exchange_weak) with `current` param as `null_ptr` however `pointer::as_ref` is also `unsafe`. It gives us a runtime guarantee however we still need to use unsafe keyword.

kaaboaye:
Looking into the issue of initialization I've realized that those modules can be also unloaded while updating. `erl_nif` exposes [unload](https://www.erlang.org/doc/man/erl_nif.html#initialization) function for that purpose. It's mainly development issue. Hot module reloads will leak the DATASET unless `upgrade ` and/or `unload` is provided.

So maybe instead of making sure that the DATASET is initialized only once we should make sure that upgrades are properly handled?

Does anyone here have any experience with how upgrades work? Does the library get new "static" memory space or is it reusing the old memory?

evnu:
> Does anyone here have any experience with how upgrades work? Does the library get new "static" memory space or is it reusing the old memory?

We have a long-standing issue w.r.t. upgrades: https://github.com/rusterlium/rustler/issues/13. Currently, upgrading is not allowed.

> So maybe instead of making sure that the DATASET is initialized only once we should make sure that upgrades are properly handled?

As upgrades are not supported right now, we can probably revisit that part later on. But maybe we need to figure out how to handle `unload`. I think this is probably not called usually, but only if an upgrade will happen?

rustler_tests/native/dynamic_load/src/lib.rs
evnu: The atomicity of `AtomicPtr` seems preferable, even though `on_load` should be called only once anyways. I think with `AtomicPtr`, we also don't need to hide the data, writing to the `AtomicPtr` does not happen by accident. @kaaboaye what do you think, should we use `AtomicPtr` or is the current version prettier?

kaaboaye: Frankly I would stick with Option. When upgrade happens it will cause memory leak instead of crash which I think will be preferable during development. Assuming there are no upgrades done in production environment.

kaaboaye: As long as load is even being called multiple times. I have no idea

kaaboaye:
>  But maybe we need to figure out how to handle unload.

I think `unsafe { DATASET = None }` should be enough. Rust should drop `Some(dataset)` after that.

evnu:
Ok, so only the test on windows is left here I think.

kaaboaye:
Added windows support, all tests are green on windows now.

evnu:
@kaaboaye thanks for all your work! Can you try to rebase one more time on master? We had some changes to the CI, which should make the tests run more reliably. With that, I think we can then merge here.

kaaboaye:
Done

evnu:
Seems that there is a merge conflict, can you resolve that?

evnu:
@philss resolved the merge conflict, thanks!

PR #415, opened 2021-12-07
Change supported Elixir and OTP versions

philss:
The idea is to restrict the versions we support to be the latest 3
versions of Elixir and OTP.
So right now we support Elixir ~> 1.11 and OTP ~> 22.

### DIFF
--- a/rustler_tests/mix.exs
+++ b/rustler_tests/mix.exs
@@ -5,7 +5,7 @@ defmodule RustlerTest.Mixfile do
     [
       app: :rustler_test,
       version: "0.0.1",
-      elixir: "~> 1.2",
+      elixir: "~> 1.11",
       compilers: [:rustler] ++ Mix.compilers(),
       build_embedded: Mix.env() == :prod,
       start_permanent: Mix.env() == :prod,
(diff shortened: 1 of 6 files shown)


hansihe (approved):
LGTM

evnu:
Currently, the README states:

> Rustler aims to support the newest three OTP versions as well as Elixir versions capable of running the supported versions of OTP.

This will need to be adapted as well if we also restrict the supported Elixir versions.

philss:
@evnu good point! I changed the text.

rustler_tests/mix.exs
```diff
@@ -5,8 +5,7 @@ defmodule RustlerTest.Mixfile do
     [
       app: :rustler_test,
       version: "0.0.1",
-      elixir: "~> 1.2",
-      compilers: [:rustler] ++ Mix.compilers(),
```
evnu, line 9: This seems unrelated to the PR?

philss: Sorry, it is not related. Reverted.

evnu (approved):

PR #416, opened 2021-12-07
Test mix rustler.new with Elixir v1.13

evnu:


philss (approved):
:shipit: 

It's OK the way we are testing this with a shell script, but I think in the future we could implement the test - for this mix task - in Elixir and test against our matrix. WDYT? I can give it a try if you agree :)

evnu:
> I think in the future we could implement the test - for this mix task - in Elixir and test against our matrix. WDYT? I can give it a try if you agree :)

Running the test with our build matrix makes sense. I think we don't need to reimplement `test.sh` in Elixir to do that, moving it to the jobs with our build matrix would suffice. But if we reimplement it, we could also run it under other platforms (e.g. windows), so that might still be beneficial. So if you are up to it, feel free to reimplement it in Elixir and move it to the build matrix.

ayrat555:
@evnu hello! when are you going to release this change?

evnu:
> hello! when are you going to release this change?

@ayrat555  The change here is only for testing in CI, did you mean to ask for another feature to be released?

ayrat555:
yes, I'm sorry. I meant this one - https://github.com/rusterlium/rustler/commit/d5d003dce8790c0e576411ed1521db2bc97a8225

evnu:
@ayrat555 v0.23 has been released with that change.

PR #425, opened 2022-01-19
Generic types

### LINKED ISSUE #424: NifTuple doesn't work with polymorphic structs
Small failing example:

```rust
use rustler::NifTuple;

#[derive(NifTuple)]
struct Foo<T> {
    t: T,
    t2: T,
}

fn main() {
}
```

```
error[E0107]: missing generics for struct `Foo`
 --> src/main.rs:4:8
  |
4 | struct Foo<T> {
  |        ^^^ expected 1 generic argument
  |
note: struct defined here, with 1 generic parameter: `T`
 --> src/main.rs:4:8
  |
4 | struct Foo<T> {
  |        ^^^ -
help: add missing generic argument
  |
4 | struct Foo<T><T> {
  |        ~~~~~~

For more information about this error, try `rustc --explain E0107`.
```

The code compiles both when I remove the `derive` clause, or when I remove `<T>` and change the `T` to a scalar type.

turion:
Fixes #424 . I'm trying to implement support for generic/polymorphic types. Can you give me some feedback whether I'm on the right track? If yes, I'll proceed with the other macros.


evnu:
>  Can you give me some feedback whether I'm on the right track?

This looks like the right approach to me. Is something missing? How does it interact with lifetimes?

turion:
> Is something missing?

I don't know whether this feature is now already sufficiently tested, or whether I should write an extra test somewhere

> How does it interact with lifetimes?

So far only on the syntactical level. I didn't implement lifetime polymorphism here because I don't understand yet how they interact with the lifetimes of the `Encoder` and `Decoder` traits. The only reason I touched them a bit is because syntactically they are declared in the same place as type parameters.

turion:
> Is something missing?

Other derive macros are missing, e.g. for structs and enums.

turion:
Ok, this should be good to go now. I added type parameters to all cases where it made sense (unit enums didn't make sense), and added appropriate tests I think.

turion:
Ping

rustler_codegen/src/context.rs
```diff
@@ -44,11 +45,37 @@ impl<'a> Context<'a> {
...
+                &type_param.ident
+            })
+            .collect::<Vec<_>>();
+
         let ident = &ast.ident;
-        let ident_with_lifetime = if has_lifetime {
-            quote! { #ident <'a> }
+
+        let lifetimes = if has_lifetime {
+            vec![quote! {'a}]
```
evnu, line 71: Is the restriction on one lifetime still necessary? It would be nice if we simply allowed an arbitrary number of lifetimes.

evnu: Maybe we could then write something like this:

```rust
let lifetimes: Vec<_> = ast.generics.lifetimes().collect();
let ident_with_generics = quote! { #ident < #(#lifetimes,)* #(#type_param_idents),* > };
```

rustler_tests/native/rustler_test/src/test_codegen.rs
```diff
@@ -85,6 +162,21 @@ pub fn untagged_enum_echo(untagged_enum: UntaggedEnum) -> UntaggedEnum {
...
+    Bool(bool),
+}
+
+#[rustler::nif]
+pub fn generic_untagged_enum_echo(
+    generic_untagged_enum: GenericUntaggedEnum<GenericMap<i32, i32>, &str>,
+) -> GenericUntaggedEnum<GenericMap<i32, i32>, &str> {
+    generic_untagged_enum
+}
+
```
evnu, line 179: A test with explicit lifetimes would be nice. Something along these lines maybe? Testing with lifetimes could be tricky, but a `'static` string might suffice.

```rust
#[derive(NifStruct)]
struct Slice<'a> {
  my_str: &'a str
}

#[rustler::nif]
pub fn from_a_string_slice() -> Slice {
  Slice { my_str: "hello world" }
}
```

evnu (commented):
I like where this is going. If we could also lift the restriction on a single lifetime, that would be great.

rustler_codegen/src/context.rs
turion: Lifetimes are harder, because the decoder trait has a lifetime as well. Note that this lifetime is set equal to the one of the datatype, if it has one. I haven't understood how it interacts with the generated code completely. When there are multiple lifetimes, we might need to alpha-rename one of them to prevent a name clash.

But I'll happily give it a go in a subsequent PR if that's fine for you!

rustler_tests/native/rustler_test/src/test_codegen.rs
turion: Do you mean an explicit lifetime together with a type parameter? I (hope I) haven't changed anything about the behaviour with lifetimes here. I'll send a short PR ahead that adds a lifetime test (since it seems there isn't one yet in `test_codegen.rs`) and rebase onto that, then we know that I didn't break them.

turion: Ok, seems like lifetimes are broken anyways, #428. This looks like an orthogonal issue to me.

turion:
Ping ;)

evnu:
> Ping @evnu ;)

Do you plan to complete #430 first and then we take a look into this PR here again?

rustler_codegen/src/nif.rs
```diff
@@ -19,6 +19,15 @@ pub fn transcoder_decorator(args: syn::AttributeArgs, fun: syn::ItemFn) -> Token
         .map(|ref n| syn::Ident::new(n, Span::call_site()))
         .unwrap_or_else(|| name.clone());
 
+    if fun.sig.generics.type_params().next().is_some() {
```
evnu, line 22: Can you give a small example where this panic would be triggered?

turion:
```rust
#[rustler::nif]
pub fn generic_echo<T>(t: T) -> T {
    t
}
```
This is a bit of a silly example, so here is a more relevant one:

```rust
// Ok, derive now works for polymorphic types
#[derive(NifTuple)]
pub struct GenericTuple<T> {
    t1: T,
    t2: T,
}

// This is fine, generic type is fully applied
#[rustler::nif]
pub fn generic_tuple_echo_usize(generic_tuple: GenericTuple<usize>) -> GenericTuple<usize> {
    generic_tuple
}

// This could type check, but rustler cannot work with it
#[rustler::nif]
pub fn generic_tuple_echo<T>(generic_tuple: GenericTuple<T>) -> GenericTuple<T> {
    generic_tuple
}
```

evnu: Thanks!

turion:
Clippy fails, but I didn't touch the failing code :confused:

evnu:
@turion can you rebase on master?

turion:
@evnu The lifetimes MR has changed so much that this PR here can basically be thrown away. I'm not even sure, maybe generic types are already handled now?

evnu:
> I'm not even sure, maybe generic types are already handled now?
[
We will have to try how well](https://github.com/rusterlium/rustler/pull/483) adds generic types, but maybe we need to test this further. Lets keep this PR closed. For #424, we will have to check if that is actually fixed by #483.

PR #429, opened 2022-02-10
Improve documentation of derive macros

turion:
This clarifies the usage of the `Nif*` derive macros a bit, and removes a few small errors.

### DIFF
--- a/rustler_codegen/src/lib.rs
+++ b/rustler_codegen/src/lib.rs
@@ -20,8 +20,8 @@ enum RustlerAttr {
     Tag(String),
 }
 
-/// Implementation of a Native Implemented Function (NIF) macro that lets the user annotate
-/// a function that will be wrapped in higer-level NIF implementation.
+/// Initialise the Native Implemented Function (NIF) environment
+/// and register NIF functions in an Elixir module.
 ///
 /// ```ignore
 /// #[rustler::nif]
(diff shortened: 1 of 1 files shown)


rustler_codegen/src/lib.rs
```diff
@@ -70,9 +70,9 @@ pub fn nif(args: TokenStream, input: TokenStream) -> TokenStream {
     nif::transcoder_decorator(args, input).into()
 }
 
-/// Implementation of the `NifStruct` macro that lets the user annotate a struct that will
-/// be translated directly from an Elixir struct to a Rust struct. For example, the following
-/// struct, annotated as such:
+/// Create a correspondence between an Elixir struct and a Rust struct.
```
hansihe, line 73: > Macro for automatically deriving `Encoder` and `Decoder` implementations which convert between a Rust struct and an Elixir struct. 

How about something like this? In my mind the word correspondence might not be the most descriptive in this case.

hansihe: The same also applies in the other cases this wording is used.

hansihe (commented):
Thanks for the PR!

Generally like the changes and think they make for a much easier to understand documentation, I just have some small nits :)

rustler_codegen/src/lib.rs
turion: That's definitely an improvement! I changed the wording further, hopefully keeping your spirit.

turion:
@hansihe What do you think about this version?

philss (approved):

PR #430, opened 2022-02-10
Dev lifetime bug

### LINKED ISSUE #428: Lifetimes seem broken
Prompted by https://github.com/rusterlium/rustler/pull/425#discussion_r803431410 I tried to add tests for lifetimes, but I couldn't do it because I got a compile error:

```
    Checking rustler_test v0.1.0 (/home/turion/Nextcloud/e/informatisch/rust/rustler/rustler_tests/native/rustler_test)
error[E0261]: use of undeclared lifetime name `'a`
   --> rustler_tests/native/rustler_test/src/test_codegen.rs:145:10
    |
145 | #[derive(NifStruct)]
    |          -^^^^^^^^
    |          |
    |          undeclared lifetime
    |          lifetime `'a` is missing in item created through this procedural macro
    |
    = note: this error originates in the derive macro `NifStruct` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0261`.
```

Whole code:

<details>

```
diff --git rustler_tests/lib/rustler_test.ex rustler_tests/lib/rustler_test.ex
index ec351e2..57e87d7 100644
--- rustler_tests/lib/rustler_test.ex
+++ rustler_tests/lib/rustler_test.ex
@@ -71,6 +71,7 @@ defmodule RustlerTest do
   def tuplestruct_echo(_), do: err()
   def newtype_record_echo(_), do: err()
   def tuplestruct_record_echo(_), do: err()
+  def string_slice_lifetime_echo(), do: err()
   def reserved_keywords_type_echo(_), do: err()
 
   def dirty_io(), do: err()
diff --git rustler_tests/native/rustler_test/src/lib.rs rustler_tests/native/rustler_test/src/lib.rs
index 402637f..808a5ac 100644
--- rustler_tests/native/rustler_test/src/lib.rs
+++ rustler_tests/native/rustler_test/src/lib.rs
@@ -67,6 +67,7 @@ rustler::init!(
         test_codegen::tuplestruct_echo,
         test_codegen::newtype_record_echo,
         test_codegen::tuplestruct_record_echo,
+        test_codegen::string_slice_lifetime,
         test_dirty::dirty_cpu,
         test_dirty::dirty_io,
         test_range::sum_range,
diff --git rustler_tests/native/rustler_test/src/test_codegen.rs rustler_tests/native/rustler_test/src/test_codegen.rs
index 748934e..b0dda3a 100644
--- rustler_tests/native/rustler_test/src/test_codegen.rs
+++ rustler_tests/native/rustler_test/src/test_codegen.rs
@@ -142,6 +142,17 @@ pub fn tuplestruct_record_echo(tuplestruct: TupleStructRecord) -> TupleStructRec
     tuplestruct
 }
 
+#[derive(NifStruct)]
+#[module = "StringSliceLifetime"]
+struct StringSliceLifetime<'a> {
+    message: &'a str,
+}
+
+#[rustler::nif]
+pub fn string_slice_lifetime() -> StringSliceLifetime<'static> {
+    StringSliceLifetime { message: "hi" }
+}
+
 pub mod reserved_keywords {
     use rustler::{NifMap, NifRecord, NifStruct, NifTuple, NifUntaggedEnum};
 
diff --git rustler_tests/test/codegen_test.exs rustler_tests/test/codegen_test.exs
index 2fe5684..855594c 100644
--- rustler_tests/test/codegen_test.exs
+++ rustler_tests/test/codegen_test.exs
@@ -21,6 +21,10 @@ defmodule TupleStructRecord do
   defrecord :tuplestruct, a: 1, b: 2, c: 3
 end
 
+defmodule StringSliceLifetime do
+  defstruct :message
+end
+
 defmodule RustlerTest.CodegenTest do
   use ExUnit.Case, async: true
 
@@ -231,6 +235,11 @@ defmodule RustlerTest.CodegenTest do
                  end
   end
 
+  test "string_slice_lifetime" do
+    value = %StringSliceLifetime{message: "hello"}
+    assert value == string_slice_lifetime_echo()
+  end
+
   test "reserved keywords" do
     assert %{override: 1} == RustlerTest.reserved_keywords_type_echo(%{override: 1})
```

</details>

turion:
Work in progress to fix #428


turion:
This only works accidentally because the name for the lifetime (`'a`) matches the one from the `Decoder` trait.

turion:
The solution for `Decoder` seems to be that every declared lifetime in the struct must be the `Decoder` lifetime. This sort of makes sense for typical use cases.

evnu:
> The solution for Decoder seems to be that every declared lifetime in the struct must be the Decoder lifetime.

So that would mean `'decoder_lifetime : 'b`? for a type `Foo<'b>`?

turion:
I've just set them actually equal for now. Maybe this is not flexible enough, I'm not sure, but it seems to work.

The missing part now are lifetimes in nif functions, but maybe these simply don't make much sense (in the same way that generic type arguments don't make sense there).

For example, there is a function `pub fn send_all<'a>(env: Env<'a>, pids: Vec<LocalPid>, msg: Term<'a>) -> Term<'a>` in the module `rustler/rustler_tests/native/rustler_test/src/test_env.rs`, and it only works because the lifetime there is called `'a`, as in the `nif` macro. So this is a bit like an undocumented feature. It would be cool if in the long run these lifetimes would at least have their own name, like `'rustler`, otherwise people might stumble over this and wonder why `'a` sometimes works, and `'b` doesn't.

turion:
I think this resolves the bug, but I'm not sure whether this is the right way to go.

rustler_codegen/src/context.rs
```diff
@@ -15,6 +15,8 @@ use super::RustlerAttr;
 pub(crate) struct Context<'a> {
     pub attrs: Vec<RustlerAttr>,
     pub ident: &'a proc_macro2::Ident,
+    pub lifetimes: Vec<Lifetime>,
+    // FIXME make method
```
turion, line 19: FIXME

rustler_codegen/src/ex_struct.rs
```diff
@@ -65,9 +65,13 @@ pub fn transcoder_decorator(ast: &syn::DeriveInput, add_exception: bool) -> Toke
...
 
 fn gen_decoder(ctx: &Context, fields: &[&Field], atoms_module_name: &Ident) -> TokenStream {
-    let struct_type = &ctx.ident_with_lifetime;
     let struct_name = ctx.ident;
     let struct_name_str = struct_name.to_string();
+    let lifetimes = &ctx.lifetimes;
+    // FIXME maybe put in context
+    let mut rustler_decoder_lifetimes = vec![];
+    rustler_decoder_lifetimes.resize(lifetimes.len(), quote! { 'a });
+    let struct_type = quote! { #struct_name < #(#rustler_decoder_lifetimes),* > };
```
turion, line 74: This is an ugly hack. Basically, every lifetime in the type is set to `<'a, 'a, 'a,...>`, which is the lifetime used further below in the decoder.

rustler_codegen/src/tuple.rs
```diff
@@ -34,10 +34,15 @@ pub fn transcoder_decorator(ast: &syn::DeriveInput) -> TokenStream {
...
 
 fn gen_decoder(ctx: &Context, fields: &[&Field]) -> TokenStream {
-    let struct_type = &ctx.ident_with_lifetime;
     let struct_name = ctx.ident;
     let struct_name_str = struct_name.to_string();
 
+    let lifetimes = &ctx.lifetimes;
+    // FIXME maybe put in context
+    let mut rustler_decoder_lifetimes = vec![];
+    rustler_decoder_lifetimes.resize(lifetimes.len(), quote! { 'a });
```
turion, line 43: I haven't renamed this yet, in other places neither

rustler_codegen/src/tuple.rs
```diff
@@ -34,10 +34,15 @@ pub fn transcoder_decorator(ast: &syn::DeriveInput) -> TokenStream {
 }
 
 fn gen_decoder(ctx: &Context, fields: &[&Field]) -> TokenStream {
-    let struct_type = &ctx.ident_with_lifetime;
     let struct_name = ctx.ident;
     let struct_name_str = struct_name.to_string();
 
+    let lifetimes = &ctx.lifetimes;
+    // FIXME maybe put in context
```
turion, line 41: FIXME (several of these)

rustler_codegen/src/tuple.rs
evnu, line 43: Why resize?

turion: I didn't know a cleverer way to create a vector of length `lifetimes.len()` containing `'a`.

evnu: `vec![quote! { 'a }, lifetimes.len()]`

rustler_codegen/src/untagged_enum.rs
```diff
@@ -94,8 +99,8 @@ fn gen_encoder(ctx: &Context, variants: &[&Variant]) -> TokenStream {
         .collect();
 
     let gen = quote! {
-        impl<'b> ::rustler::Encoder for #enum_type {
-            fn encode<'a>(&self, env: ::rustler::Env<'a>) -> ::rustler::Term<'a> {
+        impl<'__rustler_Encoder #(, #lifetimes)*> ::rustler::Encoder for #enum_type {
+            fn encode<'__rustler_encode>(&self, env: ::rustler::Env<'__rustler_encode>) -> ::rustler::Term<'__rustler_encode> {
                 match *self {
                     #(#variant_defs)*
```
turion, line 105: There are lifetimes in here as well, these need to be edited

rustler_codegen/src/map.rs
```diff
@@ -77,42 +85,29 @@ fn gen_decoder(ctx: &Context, fields: &[&Field], atoms_module_name: &Ident) -> T
         .unzip();
 
     let gen = quote! {
-        impl<'a> ::rustler::Decoder<'a> for #struct_type {
-            fn decode(term: ::rustler::Term<'a>) -> Result<Self, ::rustler::Error> {
+        impl<'__rustler_Decoder #(, #lifetimes : '__rustler_Decoder)*> ::rustler::Decoder<'__rustler_Decoder> for #struct_type {
```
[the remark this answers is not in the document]
evnu: Did you check the resulting code with `cargo expand`?

SeokminHong:
I made a basic implementation that supports generics for `NifMap` macro, based on this PR: https://github.com/rusterlium/rustler/compare/master...SeokminHong:rustler:clippy-warning

The following structure
```rust
#[derive(NifMap)]
pub struct WithLifetimes<'a, 'b> {
    pub i: rustler::Binary<'a>,
    pub j: rustler::Binary<'b>,
}
```
will be converted to :
```rust
impl<'a> ::rustler::Decoder<'a> for WithLifetimes<'a, 'a> {
    fn decode(term: ::rustler::Term<'a>) -> ::rustler::NifResult<Self> {
        ...
    }
}
impl<'a, 'b> ::rustler::Encoder for WithLifetimes<'a, 'b> {
    fn encode<'__rustler_encode>(
        &self,
        env: ::rustler::Env<'__rustler_encode>,
    ) -> ::rustler::Term<'__rustler_encode> {
        ...
    }
}
```

And I believe it works. You may notice the structure's lifetime implementing `Decoder` has `'a, 'a`, because there is a trait bound for `Decoder<'a>`, `Sized + 'a`.

@turion said this seems like a hack (https://github.com/rusterlium/rustler/pull/430#discussion_r830090544), but I think it is the right way to do it.

If the above decoder were the following code:
```rust
impl<'a, 'x: 'a, 'y: 'a> ::rustler::Decoder<'a> for WithLifetimes<'x, 'y> {
    fn decode(term: ::rustler::Term<'a>) -> ::rustler::NifResult<Self> {
        ...
        let rustler_map_field_0_i = try_decode_field::<'x>(term, atom_i())?; // error
        let rustler_map_field_1_j = try_decode_field::<'y>(term, atom_j())?; // error
        Ok(WithLifetimes {
            i: rustler_map_field_0_i,
            j: rustler_map_field_1_j,
        })
    }
}
```

the compiler complains during decoding fields because the `term` is bound to `'a`, but the `try_decode_field` expects `'x` and `'y`.

evnu:
@SeokminHong That sounds reasonable. I think we should use that approach.  @turion what do you think? @neosimsim this would also resolve #470, as lifetimes are only generated if needed.

turion:
Ok with me. I have trouble rebasing semantically though. If you have some hints on how to debug, I'd be grateful and try to proceed.

evnu:
@turion, @SeokminHong I pulled in your prior work/ideas into https://github.com/rusterlium/rustler/pull/483. I added both of you as `Co-authored-by` to the initial commit, are you ok with that? I want to give both of you credit for the work here.

turion:
Closed in favour of https://github.com/rusterlium/rustler/pull/483

PR #434, opened 2022-02-24
Bump toml

### LINKED ISSUE #433: Is it possible to upgrade toml to ~> 0.6 or loosen the version restiction?
Another dep in my project uses a more recent version of toml causing rustler to fail to be added as a dep:
```
$>  mix deps.get
Resolving Hex dependencies...                                                                                                                                                      
                                                                                                                                                                                   
Failed to use "toml" because                                                                                                                                                       
  rustler (version 0.23.0) requires ~> 0.5.2                                                                                                                                       
  mix.lock specifies 0.6.2    
```

Looks like there haven't been any removals from the public interface: https://diff.hex.pm/diff/toml/0.5.2..0.6.2

I'd be happy to open a PR for this, just unsure of y'alls process or what the best course of action is.

evnu:
This PR bumps the toml dependency of `rustler_mix` and changes the version requirement to `~> 0.6`.

Fix #433.

### DIFF
--- a/rustler_mix/mix.exs
+++ b/rustler_mix/mix.exs
@@ -24,7 +24,7 @@ defmodule Rustler.Mixfile do
 
   defp deps do
     [
-      {:toml, "~> 0.5.2", runtime: false},
+      {:toml, "~> 0.6", runtime: false},
       {:ex_doc, ">= 0.0.0", only: :dev, runtime: false},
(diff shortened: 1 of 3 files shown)


evnu:
Seems that we need to use the new version of `rustler_sys` already. Closing in favour of #435 (which includes the bump here).

PR #439, opened 2022-03-16
Abort compilation on macos if linker arguments are missing

### LINKED ISSUE #235: macOS platform hack does not pick up workspace cargo configuration

evnu:
This PR adds a mechanism to search through `.cargo/config` files up from the crate to the workspace root in order to find target configuration for macos. If any such configuration is found, we assume that it contains the linker arguments required to make rustler work on macos.

Fix #235.

### DIFF
--- a/rustler_mix/lib/rustler/compiler.ex
+++ b/rustler_mix/lib/rustler/compiler.ex
@@ -21,7 +21,8 @@ defmodule Rustler.Compiler do
         |> make_features_flag(config.features)
         |> make_target_flag(config.target)
         |> make_build_mode_flag(config.mode)
-        |> make_platform_hacks(crate_full_path, :os.type())
+
+      ensure_platform_requirements!(crate_full_path, config, :os.type())
 
       compile_result =
         System.cmd(cmd, args,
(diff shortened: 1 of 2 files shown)


filmor:
If we do that, we might as well just force the correct arguments...

evnu:
> If we do that, we might as well just force the correct arguments...

You mean as in adding them to the `args`? The problem with that approach is that we then need to call `cargo build` for each target separately, see https://github.com/rusterlium/rustler/pull/401#issuecomment-958779533.

rustler_mix/lib/rustler/compiler.ex
```diff
@@ -69,46 +70,58 @@ defmodule Rustler.Compiler do
...
+
+    unless has_macos_target_os_configuration? do
+      raise """
+      Compiling on macOS requires special link args in order to compile
+      correctly.
+
+      To remove this error, please create .cargo/config
+      with the following content:
+
+            [target.x86_64-apple-darwin]
```
philss, line 112: I think this suggestion can use the `cfg` for capturing the OS as you did in the other PR: https://github.com/rusterlium/rustler/pull/438/files#diff-aea852ba0a1ff20d4a5eab3df8c59e3fc4e91560f8a0e07d131e1ca2ca600127R93

```suggestion
            [target.'cfg(target_os = "macos")']
```

rustler_mix/lib/rustler/compiler.ex
```diff
@@ -69,46 +70,58 @@ defmodule Rustler.Compiler do
...
+      |> Enum.reverse()
+
+    has_macos_target_os_configuration? =
+      sorted_config_files
+      |> Stream.map(&Toml.decode_file!/1)
+      |> Enum.drop_while(fn toml ->
+        toml
+        |> Map.get("target", [])
+        |> Enum.filter(fn {key, _} ->
+          String.match?(key, ~r/(.*macos.*)|(.*darwin.*)/)
```
philss, line 100: Couldn't we check the args here as suggested by @filmor?

```suggestion
          String.match?(key, ~r/(.*macos.*)|(.*darwin.*)/) and String.match?(value, ~r/(.*link-arg=-undefined.*)|(.*link-arg=dynamic_lookup.*)/)
```

I'm not sure if `value` is a string, though.

philss (commented):
Some suggestions :)

rustler_mix/lib/rustler/compiler.ex
evnu: Ah, maybe I misunderstood @filmor, I thought he suggested to add the correct arguments to `args`. I think we can check the correct arguments here. I will figure out what the Toml parser returns for the key-value pairs.

rustler_mix/lib/rustler/compiler.ex
evnu: Yeah, this conflicted with #438, I will rebase on that.

rustler_mix/lib/rustler/compiler.ex
evnu: See f16cbf2..2697b62, I added some code to match the specific arguments. I think this might be a bit brittle. Maybe it is enough if the target configuration exists at all, as that implies that a developer already thought about what the correct arguments should be?

philss: Yeah, I think it would be enough to check if the target config exists :+1:

philss (approved):

evnu:
I will merge this for now. @filmor and @rusterlium/core, I am open to reverting this if someone can think of a better strategy to handle linker arguments for macos.

Issue #111, opened 2017-08-26
how to use dirty nifs


benoitc:
is there a way to pass dirty nifs flags when exporting the functions?

benoitc:
also is this needed?

hansihe:
Added some very basic tests for this to the test suite, https://github.com/hansihe/rustler/commit/99544fcd90445b4ae6e6cf1b9371d63ba5f6a541.

This should definitely be much better documented.

Issue #402, opened 2021-11-07
Consider adding a warning about lengthy work


dlesl:
Rustler is quite mature now and thanks to its great documentation it's easy to get a project working without reading the [official erl_nif docs](https://www.erlang.org/doc/man/erl_nif.html), which unfortunately means that users might not be aware of the big warning box there. While rustler can help eliminate memory errors, the warning about [lengthy work](https://www.erlang.org/doc/man/erl_nif.html#lengthy_work) is still just as valid, and for many projects, finding a way to do lengthy work correctly will end up being the most challenging part of writing a NIF. I think it would be helpful to:
- Add a warning to the rustler readme with a link to the erl_nif docs
- Link to the ways rustler supports the strategies mentioned in the erl_nif docs
  - [Threaded NIFs](https://docs.rs/rustler/0.22.2/rustler/thread/index.html)
  - [Dirty NIFs](https://docs.rs/rustler/0.22.2/rustler/schedule/enum.SchedulerFlags.html)
  - And Yielding NIFs once #232 lands

dlesl:
[enif_consume_timeslice](https://docs.rs/rustler/0.22.2/rustler/schedule/fn.consume_timeslice.html) would be good to mention too

scrogson:
Agreed.

PR #444, opened 2022-04-11
Document the usage of `schedule` flag for `nif` macro

philss:
Closes https://github.com/rusterlium/rustler/issues/111


philss:
This is also related to https://github.com/rusterlium/rustler/issues/402, but don't solves that completely.

PR #445, opened 2022-04-12
Update links in the docs and bump ex_doc

philss:
The idea is to improve documentation and make easier for users to find
things.


rustler_mix/lib/rustler.ex
```diff
@@ -75,7 +75,7 @@ defmodule Rustler do
           load_data: :something
       end
 
-  [target]: https://forge.rust-lang.org/release/platform-support.html
+  [target]: https://doc.rust-lang.org/nightly/rustc/platform-support.html
```
evnu, line 78: Why nightly?

evnu (approved):

rustler_mix/lib/rustler.ex
philss: It was the first link redirected from that old one. I changed to be `stable`.

PR #448, opened 2022-04-20
Fix compilation error of `Result` alias

### LINKED ISSUE #447: Deriving macro doesn't work with `Result` alias
If a module uses a re-defined `Result` like the following code:

```rust
type Result<T> = std::result::Result<T, MyError>;
```

When you use `Nif` macros derivation below, it may occur the following error.

```rust
#[derive(NifMap)]
struct Test {
    a: i32,
}
```

```
error[E0107]: this type alias takes 1 generic argument but 2 generic arguments were supplied
 --> src/lib.rs:8:10
  |
8 | #[derive(rustler::NifMap)]
  |          ^^^^^^^^^^^^^^^
  |          |
  |          expected 1 generic argument
  |          help: remove this generic argument
  |
note: type alias defined here, with 1 generic parameter: `T`
 --> src/lib.rs:6:6
  |
6 | type Result<T> = std::result::Result<T, MyError>;
  |      ^^^^^^ -
```

Repo: https://github.com/SeokminHong/rustler-iss-result

SeokminHong:
Fixes #447

### DIFF
--- a/rustler_codegen/src/ex_struct.rs
+++ b/rustler_codegen/src/ex_struct.rs
@@ -96,7 +96,7 @@ fn gen_decoder(ctx: &Context, fields: &[&Field], atoms_module_name: &Ident) -> T
 
     let gen = quote! {
         impl<'a> ::rustler::Decoder<'a> for #struct_type {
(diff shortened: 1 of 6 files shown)


filmor (approved):

evnu:
Thanks! :)

PR #454, opened 2022-05-15
Add support for large integers

thomas9911:
erlang / elixir support integers of arbitrary size, however the erlang nif does not support this (if i read the docs correctly), I used the term to binary to extract this data. 

So couple of questions:
- do we want to support this in Rustler?
- do we want to use num-bigint for this or another libary?
- should it be behind a feature flag?
- is the term to binary / binary to term functions the way to do this?


evnu:
Thanks @thomas9911. Using the binary representation of a term is nice, do you know how stable that is? I'd think that the representation does not change ever, but is that guaranteed?

> do we want to support this in Rustler?

I am not sure if the `rustler` crate is the right spot, maybe a `rustler_bigint` crate would expose this functionality more clearly and separate core `erl_nif` handling from extra features such as this. But if we add an extra crate, we need to add a newtype `BigInt` (as you also stated in the description). The ergonomics would be slightly worse, but retrieving the BigInt from the newtype is also trivial, so this might be ok. What do you think?

> do we want to use num-bigint for this or another libary?

I haven't used a bigint library yet with Rust, but `num-bigint` seems to be the most popular option, so this should be a good choice.

> should it be behind a feature flag?

That depends if we want this to be in a separate crate or not. If we don't use a separate crate, I think a feature flag would be good (to avoid pulling in `num-bigint` for users that don't actually need it).

> is the term to binary / binary to term functions the way to do this?

That depends on the stability of the in-memory term representation. I guess that the [external term format](https://www.erlang.org/doc/apps/erts/erl_ext_dist.html) provides more stability guarantees, but if we use that, we lose the nice property of simply passing in a term, which can then be decoded. Maybe others in @rusterlium/core have more insight into this?

thomas9911:
> I'd think that the representation does not change ever, but is that guaranteed?
After a google search I could find docs of this of erts version 5.6.3 (we are now on 13.0) and it looks quite the same (or at least backwards compatible) and exactly the same for the types i used in this pull request. So indeed no guarantees, but strong suspicion.

> I am not sure if the rustler crate is the right spot, maybe a rustler_bigint crate would expose this functionality more clearly

After creating this pull request I think indeed it is probably better to implement this in a seperate crate.

> That depends on the stability of the in-memory term representation.

I looked at the nif and this uses exactly the external term format in these functions.

evnu:
That sounds reasonable. So lets create `rustler_bigint` as part of the `rustler` repository with a newtype and use your approach with reading the binary directly?

thomas9911:
Sounds good, shall i create a new folder ect. in this pull request, or do you want to do it?

evnu:
> Sounds good, shall i create a new folder ect. in this pull request, or do you want to do it?

You can create a new crate in this pull request.

rustler_bigint/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "rustler_bigint"
+version = "0.1.0"
```
thomas9911, line 3: should this version be the same as `rustler` ? (or at least the minor version)

thomas9911:
Moved everything to rustler_bigint crate and added extra tests and documentation

rustler_bigint/Cargo.toml
evnu: Good question! I think we should use a different versioning, starting with `0.1.0`, as we probably want to indicate that this is new and not battle-tested yet. We can merge the versions when bumping rustler to `1.0`. Does that make sense?

rustler_bigint/src/big_int.rs
```diff
@@ -0,0 +1,396 @@
...
+    }
+}
+
+impl std::ops::DerefMut for BigInt {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
+
+fn decode_big_integer(input: &[u8]) -> NifResult<num_bigint::BigInt> {
```
evnu, line 81: I think we should define constants for the magical numbers here:

```rust
// From https://www.erlang.org/doc/apps/erts/erl_ext_dist.html
const SMALL_INTEGER = 97u8;
const INTEGER = 98u8;
const  SMALL_BIG_EXT = 110u8;
const LARGE_BIG_EXT = 111u8;
```

rustler_bigint/src/big_int.rs
```diff
@@ -0,0 +1,396 @@
...
+    if input[0] != 131 {
```
evnu, line 82: We should also define a constant for that. It appears this is the external term format version number, so lets use that: `const EXTERNAL_TERM_FORMAT_VERSION = 131u8`;

rustler_bigint/src/big_int.rs
```diff
@@ -0,0 +1,396 @@
...
+
+    Ok(big_int)
+}
+
+fn encode_big_integer(big_int: &num_bigint::BigInt) -> Vec<u8> {
+    let (sign, data) = big_int.to_bytes_le();
+    let sign = if sign == Sign::Minus { 1 } else { 0 };
+
+    let mut out = vec![131];
+    if data.len() < 256 {
```
evnu, line 132: Should we also encode into small integer and integer?

rustler_bigint/src/big_int.rs
```diff
@@ -0,0 +1,396 @@
...
+    };
+
+    Ok(big_int)
+}
+
+fn encode_big_integer(big_int: &num_bigint::BigInt) -> Vec<u8> {
+    let (sign, data) = big_int.to_bytes_le();
+    let sign = if sign == Sign::Minus { 1 } else { 0 };
+
+    let mut out = vec![131];
```
evnu, line 131: `EXTERNAL_TERM_FORMAT_VERSION`

rustler_bigint/src/big_int.rs
```diff
@@ -0,0 +1,396 @@
...
+    }
+}
+
+impl Encoder for BigInt {
+    fn encode<'c>(&self, env: Env<'c>) -> Term<'c> {
+        let binary = encode_big_integer(&self.0);
+        if let Some((term, _)) = env.binary_to_term(&binary) {
+            term
+        } else {
+            error().encode(env)
```
evnu, line 161: Do you know how this error could be triggered?

rustler_bigint/src/big_int.rs
evnu: Also: Do we need to allow `input[0] >= 131`? I would expect that the format will be backwards compatible.

rustler_bigint/src/big_int.rs
thomas9911: I thought about it, but we load it into erlang in the next step. And i thought maybe the logic to implement this is "more work" than the runtime overhead it takes to load a bit larger number.

I shall write a small benchmark to verify my assumptions

rustler_bigint/src/big_int.rs
thomas9911: I can't find on the internet if there ever was a different version. So maybe it is better to keep it like this and if another version comes out, re-evaluate this.

rustler_bigint/src/big_int.rs
thomas9911: Written a small benchee benchmark, which loads the number 50 if big integer format and in small integer format
```
big_int = <<131, 110, 1, 0, 50>>
small_int = <<131, 97, 50>>
int = <<131, 98, 0, 0, 0, 50>>

Benchee.run(
  %{
    "small_int" => fn -> :erlang.binary_to_term(small_int) end,
    "int" => fn -> :erlang.binary_to_term(int) end,
    "big_int" => fn -> :erlang.binary_to_term(big_int) end
  }
)

```

out:

```
Comparison:
int             15.64 M
small_int       15.21 M - 1.03x slower +1.81 ns
big_int         14.22 M - 1.10x slower +6.36 ns
```

big int is indeed a bit slower 🤔

thomas9911: good news, found an easy way to implement it, so I have added decoding as normal integer (98)

rustler_bigint/src/big_int.rs
thomas9911: done

rustler_bigint/src/big_int.rs
thomas9911: done

rustler_bigint/src/big_int.rs
```diff
@@ -0,0 +1,396 @@
...
+
+fn decode_big_integer(input: &[u8]) -> NifResult<num_bigint::BigInt> {
+    if input[0] != 131 {
+        return Err(Error::BadArg);
+    };
+
+    let big_int = match input[1] {
+        97 => {
+            // small integer / byte
+            num_bigint::BigInt::from(input[2])
```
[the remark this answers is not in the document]
thomas9911: done

rustler_bigint/src/big_int.rs
evnu: Makes sense.

rustler_bigint/src/big_int.rs
evnu: >  ran it a couple of times and the results are all over the place :D so i dont think there is a slow down between the three

Yeah, for such benchmarks, Erlang's scheduling will often result in flaky results. You have to re-run within the benchmark itself using a loop and then adjust the results, that often helps.

rustler_bigint/src/big_int.rs
evnu: I think we can/should assume that `encode_big_integer` is correct. Also, should we maybe call `encode(env)` on the `Vec<u8>`?

```rust
fn encode<'c>(&self, env: Env<'c>) -> Term<'c> {
  encode_big_integer(&self.0).encode(env)
}
```

evnu (changes requested):
I think we are almost there!

rustler_bigint/src/big_int.rs
thomas9911: We have to still convert it from etf bytes to a valid erlang/elixir term. 
Did a small refactor to return nil instead when invalid bytes are encountered (ofcourse it should never happen)

evnu: Maybe we should raise an exception if conversion fails, e.g. `::rustler::Error::RaiseTerm(format!("could not encode bigint {}", bigint))`. What do you think? I think `nil` might be quite unexpected.

thomas9911: Yes, lets do that. Found the nil thing also quite akward while I implemented it :sweat_smile:

thomas9911: Small issue, is it possible to convert an Error to a Term? I don't think we can right? (because how can you encode a raise statement in a term :thinking: )

thomas9911: updated to return an error tuple

evnu: Looks good, thanks!

evnu (approved):

joshuataylor:
This is amazing, I'm so happy this in! Thanks!

There is also a discussion here that touches on the topic above as well https://elixirforum.com/t/acceptance-of-erlangs-term-to-binary-and-vice-versa-in-elixir/11034

PR #455, opened 2022-05-25
ci: Test OTP 25 and remove OTP 22

evnu:


filmor:
Could you maybe, while you're at it, update OTP23/24 to the latest available point release?

PR #465, opened 2022-06-19
Prepare rustler_bigint release

evnu:
This PR adds a `README.md` and relevant information in `Cargo.toml` to `rustler_bigint`. With that, we should be able to release the crate soon-ish as a v0.1. @thomas9911  please take a look if I missed something, or maybe if we should add additional information to the README.

### DIFF
--- a/rustler_bigint/README.md
+++ b/rustler_bigint/README.md
@@ -1,5 +1,9 @@
 # Rustler BigInt
 
+[Documentation](https://docs.rs/rustler_bigint/latest/rustler_bigint)
+[![Build Status](https://github.com/rusterlium/rustler/workflows/CI/badge.svg?branch=master)](https://github.com/rusterlium/rustler/actions/workflows/main.yml)
+[![Crates.io package version](https://img.shields.io/crates/v/rustler_bigint.svg)](https://crates.io/crates/rustler_bigint)
+
 `rustler_bigint` provides support for Erlang's arbitrarily-sized integers.
 
 ## Installation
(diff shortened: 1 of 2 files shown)


rustler_bigint/README.md
```diff
@@ -0,0 +1,34 @@
...
+rustler_bigint = { version = "0.1" }
+```
+
+## Example
+
+Lets assume that we need to handle integers of variable size. Some might fit
+into Rust's `i64`, but others might not. For example:
+
+```elixir
+large = 2**65 # This does not fit into i64, it is an Erlang big integer
```
thomas9911, line 20: maybe use the Bitwise.bsl function because 2**65 is not supported for older elixir versions? 

```
two_to_the_65 =  Bitwise.bsl(2, 65) # This does not fit into i64, it is an Erlang big integer
```

rustler_bigint/README.md
```diff
@@ -0,0 +1,34 @@
+# Rustler BigInt
+
```
thomas9911, line 2: maybe add a link to the docs? `https://docs.rs/rustler_bigint/latest/rustler_bigint/` (or ofcourse the real url when we know what it is :P )

rustler_bigint/README.md
```diff
@@ -0,0 +1,38 @@
...
+In Rust, we can use `rustler_bigint::BigInt` to pass integer values of
+different sizes into a NIF. The type `rustler_bigint::BigInt` is a newtype
+wrapping `num_bigint::BigInt` and implements `std::ops::Deref`, so functions
+from `num_bigint::BigInt` can be called directly.
+
+```rust
+/// Simply echo `large` back to the caller.
+#[rustler::nif]
+pub fn handle_large(large: rustler_bigint::BigInt) -> NifResult<BigInt> {
```
thomas9911, line 35: Was looking at this again, seeing:
```suggestion
pub fn handle_large(large: rustler_bigint::BigInt) -> NifResult<rustler_bigint::BigInt> {
```

evnu: Ah, good catch, thanks.

thomas9911:
Looks good to me :+1:

evnu:
@thomas9911 I just released `rustler_bigint`. Thanks for your work! I also sent you an invite to be an owner of `rustler_bigint` on `crates.io`.

PR #471, opened 2022-07-04
Remove lifetime from Encoders

### LINKED ISSUE #470: clippy complains about unused liftetimes with rust 1.62.0
After updating to rust 1.62.0 clippy complains, when using `NifUntaggedEnum`.

```
error: this lifetime isn't used in the impl
   |
23 | #[derive(Debug, NifUntaggedEnum, Clone, Copy, PartialEq)]
   |                 ^^^^^^^^^^^^^^^
   |
   = note: `-D clippy::extra-unused-lifetimes` implied by `-D warnings`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#extra_unused_lifetimes
   = note: this error originates in the derive macro `NifUntaggedEnum` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Here is a minimal example `main.rs`

```rust
use rustler_codegen::NifUntaggedEnum;

#[derive(Debug, NifUntaggedEnum, Clone, Copy, PartialEq)]
pub enum Timestamp {
    Microseconds(i64),
}


fn main() {
    println!("Hello, world!");
}
```

neosimsim:
Closes #470

### DIFF
--- a/rustler_codegen/src/ex_struct.rs
+++ b/rustler_codegen/src/ex_struct.rs
@@ -163,7 +163,7 @@ fn gen_encoder(
     };
 
     let gen = quote! {
-        impl<'b> ::rustler::Encoder for #struct_type {
+        impl ::rustler::Encoder for #struct_type {
             fn encode<'a>(&self, env: ::rustler::Env<'a>) -> ::rustler::Term<'a> {
                 use #atoms_module_name::*;
                 let mut map = ::rustler::types::map::map_new(env);
--- a/rustler_codegen/src/map.rs
+++ b/rustler_codegen/src/map.rs
@@ -127,7 +127,7 @@ fn gen_encoder(ctx: &Context, fields: &[&Field], atoms_module_name: &Ident) -> T
         .collect();
 
     let gen = quote! {
-        impl<'b> ::rustler::Encoder for #struct_type {
+        impl ::rustler::Encoder for #struct_type {
             fn encode<'a>(&self, env: ::rustler::Env<'a>) -> ::rustler::Term<'a> {
                 use #atoms_module_name::*;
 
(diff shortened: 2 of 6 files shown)


neosimsim:
I removed the lifetimes because clippy complains and I'm unable to see how they are supposed to work anyway.

evnu:
Thanks for the PR. I think we need to fix lifetimes for good, see https://github.com/rusterlium/rustler/issues/428 as well.

> What effect is the lifetime in Encoder supposed to have?

The lifetime should rename a pre-existing lifetime for a struct, if a struct has a lifetime. When `rustler_codegen` was introduced with 594afed3, handling a struct with an associated lifetime was handled with this:

```rust
pub fn gen_encoder(struct_name: &Ident, fields: &[Field], atom_defs: &Tokens, has_lifetime: bool) -> Tokens {
    let struct_type = if has_lifetime {
        quote! { #struct_name <'b> }
    } else {
        quote! { #struct_name }
    };
// ...
    quote! {
        impl<'b> rustler::NifEncoder for #struct_type {
// ...
}
```

So back then, the lifetime was renamed to `'b`. We refactored that part in 6be843b6:

```rust
    let struct_type = &ctx.ident_with_lifetime;
```

The `ident_with_lifetime` is defined in `context.rs`:

```rust
        let ident_with_lifetime = if has_lifetime {
            quote! { #ident <'a> }
        } else {
            quote! { #ident }
        };
```

Now, I that code is wrong: the lifetime needs to be named `'a` for the `Decoder`, but `'b` for the `Encoder`. This is what caused https://github.com/rusterlium/rustler/issues/428.

SeokminHong:
@evnu It sounds to me the codegen only can handle exactly one or zero lifetime for structs, right?

https://github.com/rusterlium/rustler/blob/f946092552f8057dabdb13c66fb9364ce03120a3/rustler_codegen/src/context.rs#L41-L45

But it may be good if Rustler supports more flexible lifetimes for them. I think The approach of #430 may solve this problem by using *ugly* names for lifetimes for Rustler, but I cannot be sure is it active...

PR #473, opened 2022-07-07
Make map_from_arrays and map_from_pairs generic

dvic:
This PR makes the `map_from_arrays` and `map_from_pairs` accept values that can be turned into an iterator. This avoids additional allocations required to have a slice when calling these functions.

### DIFF
--- a/rustler/src/types/binary.rs
+++ b/rustler/src/types/binary.rs
@@ -102,7 +102,7 @@ use std::{
 /// See [module-level doc](index.html) for more information.
 pub struct OwnedBinary(ErlNifBinary);
 
-impl<'a> OwnedBinary {
+impl OwnedBinary {
     pub unsafe fn from_raw(inner: ErlNifBinary) -> OwnedBinary {
         OwnedBinary(inner)
     }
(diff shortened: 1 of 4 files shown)


dvic:
Should be good now 👍

dvic:
Not sure why the clippy started complaining about the lifetimes in `binary.rs` and `list.rs`, fixed that by removing those lifetimes.

evnu:
Thanks for the PR! Do you maybe have a benchmark how much this saves?

dvic:
> Thanks for the PR! Do you maybe have a benchmark how much this saves?

Thanks for asking because I have no idea why, but this PR is 10x slower 😀, I'll have to investigate...

dvic:
@evnu I figured it out but now the "correct" implementation is a breaking change. It comes down to `&(A, B)` vs `(&A, &B)`. When I make the iterator type `(&A, &B)`, the performance is the same but I can't pass `&a` where `let a: Vec<(A, B)>` anymore, which is a breaking change. For now I'm going to close this PR as there is no increased performance benefit and then it doesn't justify a breaking change I think.

PR #475, opened 2022-07-10
Add inline tags to commonly used functions

joshuataylor:
I've gone through and added inline to some commonly used functions.

If everyone is happy, we could also add these to others.

We should judge **what** should be inlined. See https://nnethercote.github.io/perf-book/inlining.html

`None. The compiler will decide itself if the function should be inlined. This will depend on the optimization level, the size of the function, etc. If you are not using link-time optimization, functions will never be inlined across crates.`
`#[inline]. This suggests that the function should be inlined, including across crate boundaries.`

Amazingly I saw great wins with just this, as I have to call some of these functions a lot when encoding a lot of data.

This is using lto:
before
```
Operating System: macOS
CPU Information: Apple M1
Number of Available Cores: 8
Available memory: 16 GB
Elixir 1.13.4
Erlang 25.0.2

Benchmarking encode column ...

Name                    ips        average  deviation         median         99th %
encode columns         13.86       72.17 ms    ±21.40%       74.47 ms      114.95 ms
```

after:
```
Name                    ips        average  deviation         median         99th %
encode columns         15.12       66.13 ms    ±23.48%       68.36 ms      108.11 ms
```

### DIFF
--- a/rustler/src/env.rs
+++ b/rustler/src/env.rs
@@ -48,6 +48,7 @@ pub struct SendError;
 
 impl<'a> Env<'a> {
     #[doc(hidden)]
+    #[inline]
     pub(crate) unsafe fn new_internal<T>(
         _lifetime_marker: &'a T,
         env: NIF_ENV,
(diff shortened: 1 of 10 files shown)


evnu:
@joshuataylor thanks for the PR. Can you share the code for your benchmark?

rustler_codegen/src/lib.rs
```diff
@@ -118,6 +118,7 @@ pub fn nif(args: TokenStream, input: TokenStream) -> TokenStream {
 /// such that you can use the Elixir struct definition for it.
 
 #[proc_macro_derive(NifStruct, attributes(module, rustler))]
+#[inline]
```
filmor, line 121: This is run at compile time, I don't see a point in inlining it.

hansihe:
This looks like a very good idea to me.

evnu:
Interesting. I would not expect `inline` to make a real difference. Can we explain why we see a difference in the performance?

PR #482, opened 2022-07-26
Tagged enum usability & performance improvements

cleaton:
1. Support multiple units with same field names
2. Move atom / tuple decode outside the type check if, to remove duplicate decode calls

Enum with 22 cases:
before:
tagged_enum_decode          1.56 M      642.48 ns  ±1292.75%         634 ns         675 ns
after:
tagged_enum_decode          3.17 M      314.98 ns  ±2092.02%         311 ns         343 ns

(~300ns is also the tagged_enum_encode latency on my machine)

### DIFF
--- a/rustler_codegen/src/tagged_enum.rs
+++ b/rustler_codegen/src/tagged_enum.rs
@@ -2,6 +2,7 @@ use proc_macro2::{Span, TokenStream};
 use quote::{quote, quote_spanned};
 
 use heck::ToSnakeCase;
+use std::collections::HashMap;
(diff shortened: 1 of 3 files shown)


evnu:
Thanks for the PR @cleaton! I will need some time to review this as I am on vacation :)

cleaton:
> Thanks for the PR @cleaton! I will need some time to review this as I am on vacation :)

no problem, enjoy :)

evnu (approved):
Looks good to me, thanks!

PR #488, opened 2022-09-12
do not store all of `ErlNifTerm` in `Binary` struct

hansihe:
This change enables us to not call `enif_inspect_binary` when creating subbinaries.

### DIFF
--- a/rustler/src/types/binary.rs
+++ b/rustler/src/types/binary.rs
@@ -86,6 +86,7 @@
 //! [`OwnedBinary`]: struct.OwnedBinary.html
 
 use crate::{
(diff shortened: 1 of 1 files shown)


filmor:
I rebased #487 on this one, is there a particular reason why you used `*mut u8` instead of `*const u8` here?

hansihe:
@filmor I did it to avoid the cast when getting the pointer from the ErlNifBinary. 

In retrospect it is probably a good idea to do the cast there in order to have the type be more descriptive in the struct.

hansihe:
Included in https://github.com/rusterlium/rustler/pull/487

PR #489, opened 2022-09-12
Add benchmarks and use make_map_from_arrays for NifStruct

evnu:
This PR adds a mix project `rustler_benchmarks`, which currently includes a benchmark for NifStruct and one for NifRecord. I implemented the benchmarks to evaluate how changing `NifStruct` to use `make_map_from_arrays()` affects performance.


evnu:
I am not sure if this is actually useful for rustler, but maybe having benchmarks as part of the repository would be nice. Alas, `make_map_from_arrays()` does not improve performance of encoding into a struct. Sidenote: when compiling in `:release` mode, `NifRecord` blows `NifStruct` out of the water:

```
$ mix run -e Benchmark.NifStruct.run
Name                        ips        average  deviation         median         99th %
decode                 404.50 K        2.47 μs   ±651.07%        1.63 μs        7.23 μs
decode and encode      117.27 K        8.53 μs   ±306.83%        6.34 μs       33.76 μs

Comparison:
decode                 404.50 K
decode and encode      117.27 K - 3.45x slower +6.06 μs

$ mix run -e Benchmark.NifRecord.run
Name                        ips        average  deviation         median         99th %
decode                   1.20 M        0.83 μs   ±689.21%        0.49 μs        2.22 μs
decode and encode        0.80 M        1.25 μs  ±1472.17%        0.65 μs        3.51 μs

Comparison:
decode                   1.20 M
decode and encode        0.80 M - 1.50x slower +0.42 μs
```

evnu:
CC @turion You might be interested in the benchmark results [here](https://github.com/rusterlium/rustler/pull/489#issuecomment-1243942455).

evnu:
Closing for now, we can revisit this later when it is actually needed.

josevalim:
@evnu I am curious why you didn't go ahead with the struct changes? They seem like an improvement, no? Thank you. :)

evnu:
@josevalim I was hoping to see some performance benefits from the changes, but did not. The changes to the code might make things a little prettier, but maybe not by much. @rusterlium/core thoughts? I could also split this into two PRs, one for the benchmarks and one for the changes to the encoder.

scrogson:
Having the benchmark app in the project allows us to add and measure changes easier, so I think it's a nice to have 👍

evnu:
I extracted the benchmark code into a separate PR: https://github.com/rusterlium/rustler/pull/495 What about the changes to `ex_struct`, should we also keep those?

PR #491, opened 2022-09-12
Cache Mix.env() to trigger recompilation

### LINKED ISSUE #479: Missing recompilation when changing MIX_ENV
I'm having trouble making rustler compile my crate in `release` mode.
`MIX_ENV` does not seem to have any impact. Is there a switch I can use somewhere?

On version 0.25.0

evnu:
When using MIX_ENV to configure the build profile, changing `MIX_ENV` in a subsequent run needs to trigger the mechanism for recompilation to determine if that mode was build. For this, we need to cache value of `Mix.env()` and detect when that changes. If it changes, we write the cache, `@external_resource` can pick that up.

Close #479

### DIFF
--- a/rustler_mix/README.md
+++ b/rustler_mix/README.md
@@ -20,6 +20,17 @@ def deps do
 end
 ```
 
+In order to also allow picking the build profile with `MIX_ENV`, add the `:rustler` compiler to your `project/0` in `mix.exs`:
+
+```elixir
+def project do
+  [
+  #...
+  compilers: [:rustler | Mix.compilers()]
+  ]
+end
+```
+
 ## Usage
 
 1.  Fetch all necessary dependencies:
--- a/rustler_mix/lib/mix/tasks/compile.rustler.ex
+++ b/rustler_mix/lib/mix/tasks/compile.rustler.ex
@@ -4,16 +4,32 @@ defmodule Mix.Tasks.Compile.Rustler do
   use Mix.Task
 
   def run(_args) do
-    IO.puts([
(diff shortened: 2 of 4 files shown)


evnu:
See also https://github.com/rusterlium/rustler/issues/479#issuecomment-1244239121.

hansihe:
I haven't looked into this much, but would making rust use a different target directory for each mix compilation mode be another option?

What are the pros/cons of these two options?

evnu:
> I haven't looked into this much, but would making rust use a different target directory for each mix compilation mode be another option?

The library which is loaded is hard-coded into the module at compile time, see [the Rustler module](https://github.com/rusterlium/rustler/blob/263a02759d6ac469f58170f56a9d8f6246a355bd/rustler_mix/lib/rustler.ex#L90) (`config.lib` defaults to `true` in [Rustler.Compiler.Config](https://github.com/rusterlium/rustler/blob/263a02759d6ac469f58170f56a9d8f6246a355bd/rustler_mix/lib/rustler/compiler/config.ex#L20)). So, unless I am missing something, the library path will not change, unless a recompilation is triggered. Also, the library directory already depends on the build profile. It defaults to `_build/rustler/{debug,release}`.

So, we need to find a way to ensure that a module recompilation is triggered (`use` needs to be expanded again), if the mix environment changes.

But maybe we really should get rid of `MIX_ENV` handling altogether instead and tell users to pick the build profile explicitly in `use Rustler`.

evnu:
Closing in favor of https://github.com/rusterlium/rustler/pull/496.

PR #493, opened 2022-09-13
Make OwnedBinary Sync

### LINKED ISSUE #469: impl Sync for OwnedBinary?
I have the need to share owned binaries across threads via RwLock. My hunch is that it's safe to implement Sync for OwnedBinary, but I haven't looked at these guts in quite some time.

filmor:
Make `OwnedBinary` `Sync`, fixes #469.


hansihe (approved):
Looks good to me, should be safe.

PR #496, opened 2022-09-26
Ignore MIX_ENV for build profile

### LINKED ISSUE #479: Missing recompilation when changing MIX_ENV

evnu:
Using `MIX_ENV` to determine the build profile profile does not work well as a change to `MIX_ENV` does not necessarily result in a module recompilation. With that, a shared lib built with the wrong profile might be used. To avoid that, `MIX_ENV` support is dropped altogether in favour of defaulting to `:release` if the user does not pick a profile explicitly.

Close #479

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 See [`UPGRADE.md`](./UPGRADE.md) for additional help when upgrading to newer versions.
 
+## [unreleased]
+
+### BREAKING
+
+* `MIX_ENV` is no longer considered for determining the build profile. Now, the
+  profile defaults to `:release`. Use the `:mode` option to pick another profile explicitly.
+
 ## [0.26.0] - 2022-09-02
 
 ### Highlight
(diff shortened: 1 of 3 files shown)


rustler_tests/test/binary_test.exs
```diff
@@ -3,7 +3,7 @@ defmodule RustlerTest.BinaryTest do
 
   test "subbinary creation" do
     assert RustlerTest.make_shorter_subbinary("test") == "es"
-    assert_raise ErlangError, fn -> RustlerTest.make_shorter_subbinary("t") end
```
evnu, line 6: @hansihe the assertion tested that the function `make_shorter_subbinary()` panics due to a `usize` underflow. With profile `release`, this panic does not happen any longer: the `length` wraps around, causing `make_subbinary()` to return an error. Is this OK, or should the function `make_shorter_subbinary()` panic (unwrap on `checked_sub()`)?

evnu:
I will merge this. Using `:release` mode should be a sane default, and it can be overridden easily.

PR #501, opened 2022-11-30
rustler_codegen: Fix stale doc example for init

ishitatsuyuki:
`Some(load)` has been replaced with `load = load`.

Also add a stub for showing how `load` should look like.


evnu:
Thanks!

PR #502, opened 2022-11-30
ci: Add Elixir v1.14 and fix rustler_mix_test job

evnu:


evnu:
Something broke with `setup-beam`:

> Error: Requested Erlang/OTP version (23) not found in version list (should you be using option 'version-type': 'strict'?)

evnu:
The function  to retrieve the list of prebuilt OTP versions is [here](https://github.com/erlef/setup-beam/blob/b980a5ac9b7316a23ebc16de2bb69b7c3ad731b5/src/setup-beam.js#L222). Accessing the URLs manually with a browser is denied, maybe that's the case for the setup action as well.

filmor:
See here: https://github.com/erlef/setup-beam#compatibility-between-operating-system-and-erlangotp

`ubuntu-latest` will not work with OTP23, you'll need to change that to `ubuntu-20.04`.

filmor (approved):

PR #505, opened 2022-12-09
Allow arbitrary NIF names (e.g. those ending with `!` or `?`)

KoviRobi:
There is as far as I know, no reason for the erlang function name to be an identifier.

To use, do

```rust
\#[nif(name = "foo!")]
fn foo_bang(...) -> ... {
    ...
}
```

### DIFF
--- a/rustler_codegen/src/nif.rs
+++ b/rustler_codegen/src/nif.rs
@@ -17,6 +17,12 @@ pub fn transcoder_decorator(args: syn::AttributeArgs, fun: syn::ItemFn) -> Token
     let argument_names = create_function_params(inputs.clone());
     let erl_func_name = extract_attr_value(args, "name").unwrap_or_else(|| name.to_string());
 
+    for chr in erl_func_name.chars() {
+        if chr > '\x7f' {
+            panic!("NIF names need to be Latin-1 encoded")
+        }
+    }
+
     quote! {
         #[allow(non_camel_case_types)]
         pub struct #name;
--- a/rustler_codegen/src/nif.rs
+++ b/rustler_codegen/src/nif.rs
@@ -15,16 +15,14 @@ pub fn transcoder_decorator(args: syn::AttributeArgs, fun: syn::ItemFn) -> Token
     let arity = arity(inputs.clone());
     let decoded_terms = extract_inputs(inputs.clone());
     let argument_names = create_function_params(inputs.clone());
-    let erl_func_name = extract_attr_value(args, "name")
-        .map(|ref n| syn::Ident::new(n, Span::call_site()))
-        .unwrap_or_else(|| name.clone());
+    let erl_func_name = extract_attr_value(args, "name").unwrap_or_else(|| name.to_string());
 
     quote! {
         #[allow(non_camel_case_types)]
         pub struct #name;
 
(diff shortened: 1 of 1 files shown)


rustler_codegen/src/nif.rs
```diff
@@ -15,16 +15,14 @@ pub fn transcoder_decorator(args: syn::AttributeArgs, fun: syn::ItemFn) -> Token
     let arity = arity(inputs.clone());
     let decoded_terms = extract_inputs(inputs.clone());
     let argument_names = create_function_params(inputs.clone());
-    let erl_func_name = extract_attr_value(args, "name")
-        .map(|ref n| syn::Ident::new(n, Span::call_site()))
-        .unwrap_or_else(|| name.clone());
+    let erl_func_name = extract_attr_value(args, "name").unwrap_or_else(|| name.to_string());
```
filmor, line 18: From looking at the code, the function name has to be Latin1 encoded (and that also limits the allowed characters).

KoviRobi:
I'm not saying using unicode isn't cumbersome, but since functions and variables in Elixir are just atoms, and atoms allow anything strings do, you can do
```elixir
defmodule Test do
  @cheese_var Macro.var(:"🧀s", __MODULE__)
  @doc """
  Hex needs cheese for the mouse
  """
  def unquote(:"🧀")(unquote(@cheese_var)) do
    IO.inspect(unquote(@cheese_var), label: "Trying to fetch cheese #")

    if unquote(@cheese_var) > 2 do
      {:error, "Out of cheese error! Redo from start."}
    else
      :ok
    end
  end
end

IO.inspect(apply(Test, :"🧀", [1]), label: "Getting 🧀")
```

rustler_codegen/src/nif.rs
KoviRobi: I'm not saying using unicode isn't cumbersome, but since functions and variables in Elixir are just atoms, and atoms allow anything strings do, you can do
```elixir
defmodule Test do
  @cheese_var Macro.var(:"🧀s", __MODULE__)
  @doc """
  Hex needs cheese for the mouse
  """
  def unquote(:"🧀")(unquote(@cheese_var)) do
    IO.inspect(unquote(@cheese_var), label: "Trying to fetch cheese #")

    if unquote(@cheese_var) > 2 do
      {:error, "Out of cheese error! Redo from start."}
    else
      :ok
    end
  end
end

IO.inspect(apply(Test, :"🧀", [1]), label: "Getting 🧀")
```

filmor: Did you test this? I checked the `erl_nif.c` and it explicitly uses latin1 atoms as far as I can see.

KoviRobi: I've added a check for it now :)

KoviRobi:
I'm not sure I understand the CI job failures -- they only say "this job failed" with no log/trace or anything

evnu:
@filmor do we still consider this PR?

filmor:
I included the relevant portion (not forcing identifier rules and checking for ASCII-ness) in #592 (will move that part to a separate PR once done). This one can be closed.

@KoviRobi I'll refer to you in the new PR, thank you for your contribution :)

Issue #2, opened 2016-01-06
Revise NifError


hansihe:
I am not happy with the way NifError works at the moment. It should be revised/improved and integrated with already written code.

scrogson:
@hansihe do you have anything in mind for this?

hansihe:
@scrogson I have not thought about it very much, but I do not think the way it currently works is very nice.

scrogson:
I think it might be time to pull in the `failure` crate and provide errors as structs. All `Result`s become: `Result<T, failure::Error>`.

PR #507, opened 2022-12-10
rustler: add a Rust type `ErlOption<T>`

### LINKED ISSUE #500: An optional crate feature to switch from Elixir `nil` to Erlang `undefined`?
First of all, thank you for creating Rustler! We have been using it since 2019 for Erlang projects, and it is working great.

We are actually using a fork of Rustler because we wanted to use atom `undefined` instead of `nil`. We wanted Rultler to encode Rust `None` into Erlang `undefined`.

`undefined` is commonly used in Erlang language and libraries to indicate absence of a value.

For example, if you create an Erlang record (which is just a tuple) without giving values for some fields, these fields are filled with `undefined`.

```erlang
%% File: example.hrl
-record(rec, {field1, field2}).
```

```console
$ erl
Erlang/OTP 25 [erts-13.1.2] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]

Eshell V13.1.2  (abort with ^G)
1> rr("example.hrl").
[rec]
2> #rec{field1 = 10}.
#rec{field1 = 10,field2 = undefined}
3> 
```

Another example is that there are many functions in `erlang` module that return `undefined` if the value is not found. For example, `erlang:process_info/2` returns `undefined` if the process is not alive ([doc](https://www.erlang.org/doc/man/erlang.html#process_info-2)).

In our fork of Rustler, we directly replaced all occurrences of `nil` with `undefined` ([diff](https://github.com/rusterlium/rustler/compare/master...tatsuya6502:rustler:erl-style-encoder-current)). It was a quick hack, but worked for us.

Now we are thinking of contributing this feature to the upstream Rustler. We are thinking of adding a crate feature something like `erlang-style-encoder` to switch from `nil` to `undefined`. When the feature is enabled, Rustler will encode Rust `None` into Erlang `undefined`.

Does this make sense? Is there any better way to do this? Any feedback is welcome.

tatsuya6502:
This PR fixes #500 by adding a Rust type`ErlOption<T>` to the Rustler crate.

`ErlOption<T>` is a wrapper around `Option<T>`. It implements Rustler's `Encoder` and `Decoder`, and serializes `None` into `undefined` atom instead of `nil`. The `undefined` atom is commonly used in Erlang libraries to express the absence of a value.

For convenience, `ErlOption<T>` implements some traits from `std`, so that it will be easy to get a reference of enclosing `Option<T>` or convert from/to `Option<T>`. Here are the examples from its doc:

```rust
use rustler::ErlOption;

// Create new `ErlOption<i32>` values via convenient functions.
let _ = ErlOption::some(1); // Wraps `Some(1)`.
let _ = ErlOption::<i32>::none();

// Convert Option<i32> values to ErlOption<i32> values.
let _ = ErlOption::from(Some(2));
let _: ErlOption<_> = Some(3).into();
let _: ErlOption<i32> = None.into();

// Get a reference of enclosing Option<T> from an ErlOption<T>.
let _: &Option<i32> = ErlOption::some(4).as_ref();

// Get a mutable reference of enclosing Option<T> from an ErlOption<T>.
let _: &mut Option<i32> = ErlOption::some(5).as_mut();

// Convert an ErlOption<i32> value to an Option<i32> value.
let _: Option<i32> = ErlOption::some(6).into();

// Compare ErlOption<T> with Option<T>.
assert_eq!(ErlOption::some(7), Some(7));
assert!(ErlOption::some(8) > Some(7));

// Call Option<T>'s methods on an ErlOption<T> via Deref and DerefMut.
assert!(ErlOption::some(9).is_some());
assert_eq!(ErlOption::some(10).unwrap(), 10);
assert_eq!(ErlOption::some(12).map(|v| v + 1), ErlOption::some(13));
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ See [`UPGRADE.md`](./UPGRADE.md) for additional help when upgrading to newer ver
 
 ## [unreleased]
 
+## Added
+
+* `ErlOption<T>` to provide an ergonomic option type for Erlang (#507, thanks @tatsuya6502)
+
 ### Changed
 
 * Use Cargo features to define the NIF version level (#537), deprecating
--- a/rustler/src/lib.rs
+++ b/rustler/src/lib.rs
@@ -38,7 +38,8 @@ mod term;
 
 pub use crate::term::Term;
 pub use crate::types::{
-    Atom, Binary, Decoder, Encoder, ListIterator, LocalPid, MapIterator, NewBinary, OwnedBinary,
+    Atom, Binary, Decoder, Encoder, ErlOption, ListIterator, LocalPid, MapIterator, NewBinary,
+    OwnedBinary,
 };
 pub mod resource;
 pub use crate::resource::ResourceArc;
--- a/rustler/src/types/atom.rs
+++ b/rustler/src/types/atom.rs
@@ -271,6 +271,10 @@ atoms! {
     /// The `nil` atom.
     nil,
 
+    /// The `undefined` atom, commonly used in Erlang libraries to express the
+    /// absence of value.
+    undefined,
+
     /// The `ok` atom, commonly used in success tuples.
     ok,
 
--- a/rustler/src/types/erlang_option.rs
+++ b/rustler/src/types/erlang_option.rs
@@ -1,0 +1,216 @@
+use super::atom;
+use crate::{Decoder, Encoder, Env, Error, NifResult, Term};
+
+use std::ops::{Deref, DerefMut};
+
+/// A wrapper type for [`Option<T>`][option] to provide Erlang style encoding. It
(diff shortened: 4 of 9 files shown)


filmor:
Maybe, instead of incorporating many small adjustments to the current error-handling, we could try to come up with a system that allows for all of these in one go, e.g. by extending the `nif` macro. I don't have a proposal for that, yet. @hansihe Do you remember what you had in mind for https://github.com/rusterlium/rustler/issues/2? :)

tatsuya6502:
ping @filmor @hansihe

tatsuya6502:
@filmor @hansihe — Can you please decide and tell me what I should do with this PR? Thanks!

tatsuya6502:
ping @filmor @hansihe

PR #523, opened 2023-03-20
Simplify CI workflow and update versions

philss:
It removes Elixir 1.11 that is not supported anymore.

It updates the version of Actions in the workflow, and also simplifies by merging parts that are common and can work together.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,29 +10,11 @@ on:
     - cron: '0 0 * * 3'
 
 jobs:
-  clippy:
-    name: Clippy ${{matrix.os}}
-    runs-on: ${{matrix.os}}-latest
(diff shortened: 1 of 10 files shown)


philss:
Now that is passing :tada: , I'm going to change the Action responsible for installing Rust to a more simple and well maintained project: https://github.com/dtolnay/rust-toolchain

I can revert later if you don't agree with that :)

philss:
image

Now it runs without warnings and takes about ~ 11 minutes after removing some pairs of OTP/Elixir. WDYT?

evnu (approved):

philss:
Thanks for reviewing, @evnu!

I'm merging this. I can revert or fix anything that you don't agree later.

PR #531, opened 2023-04-24
Use non-captured identifier in format string

evnu:
To ensure that MSRV of 1.56.1 still holds, replace the captured identifier in format strings. Capturing identifierts for format strings was introduced with Rust 1.58.0.


evnu:
I checked the MSRV with `cargo-msrv`:

```
$ cargo msrv
......
Finished The MSRV is: 1.56.1
```

philss (approved):
🚢

PR #532, opened 2023-04-24
Prepare release 0.28.0

evnu:
This PR prepares the release of 0.28.0.


philss (commented):
I think this is it. I can't check the full diff right now, but looks good to me :)

PS: I think we are missing a tag or gh release for 0.27

evnu:
@philss 0.27 is tagged as 80f6c4b926ce15dea4fac5389a67decdc9732ba8. But yes, we are missing a GH release there. I don't know how to make such a release, we should extend `prepare_release.sh` with information on how to do that I guess.

PR #536, opened 2023-05-14
Make rustler_sys forward compatible

filmor:
Makes `rustler_sys` forward compatible by falling back to the highest available version as long as the NIF major version matches.

### DIFF
--- a/rustler_sys/build.rs
+++ b/rustler_sys/build.rs
@@ -14,7 +14,8 @@ use std::process::Command;
 use std::{env, fs};
 
 const SNIPPET_NAME: &str = "nif_api.snippet";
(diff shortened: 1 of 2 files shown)


philss (approved):
LGTM! :+1: 

I think we could also update the CI to run against the latest RC version. WDYT?
For what I checked, the latest one is [`26.0-rc3`](https://github.com/erlang/otp/releases/tag/OTP-26.0-rc3),

filmor:
If #537 is agreed on, we don't need to merge this one. If it requires more discussion, we can merge this and publish it as `rustler-sys-2.3.1` tomorrow to have things working reasonably well when OTP26 comes out.

PR #537, opened 2023-05-14
Use features for minimal NIF version

filmor:
- Make rustler_sys forward compatible
- Use Cargo features to define the minimal required NIF version
- Stop using RUSTLER_NIF_VERSION in rustler_mix

Another side-effect of this change is that I have completely removed the Erlang call from the build process now as we do not attempt to auto-discover the highest available NIF version anymore. Instead, we compile by default for NIF 2.14, if anything higher is required, it has to be given explicitly as either `RUSTLER_NIF_VERSION` or by defining the respective feature.

I would suggest that we eventually (with rustler-0.29, rustler-sys-2.3.0) retire `RUSTLER_NIF_VERSION` in favour of users defining their `rustler` (or `rustler-sys`) dependency with the minimal required NIF version. This has also the huge advantage of being able to handle situations where `rustler` is used in a dependency gracefully.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ See [`UPGRADE.md`](./UPGRADE.md) for additional help when upgrading to newer ver
 
 ## [unreleased]
 
+### Changed
+
+* Use Cargo features to define the NIF version level (#537), deprecating
+  `RUSTLER_NIF_VERSION`
+
 ## [0.28.0] - 2023-04-24
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -57,14 +57,18 @@ Rustler aims to support the newest three major OTP versions as well as newest th
 
 #### Supported NIF version
 
-Rustler uses `erlang:system_info(nif_version)` to detect the supported NIF version of the Erlang/OTP
-system for which the NIF is to be compiled. It is possible to restrict the NIF version to an older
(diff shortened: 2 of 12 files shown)


philss (approved):
It makes totally sense to me! :+1: 

I think this way is much easier to control features based on the NIF version. My question is if we really need to deprecate the env var support.

My concern is that it's much easier today to set the NIF version by using the env var in the context of a precompilation workflow (we do it [here](https://github.com/philss/rustler-precompiled-action/blob/a08977cb9b49a7b53ec3b2bddb003d00344d91bf/action.yml#L57)).

The way using only features may lead to the need of defining custom features - in the user projects - to activate the correct `nif_version_*_*` on rustler, if the user is planning to support multiple NIF versions, right?

filmor:
I'm not sure I get what you are concerned about. If a library itself supports multiple NIF versions (making use of newer functions if available), it should also put this support behind feature flags. The environment variable makes this handling much more error prone and leads to binaries that are needlessly xompiled to higher NIF version levels.

It's actually a simplification of the precompilation as in almost all cases it will suffice to provide a single binary to support all relevant BEAM versions.

philss:
> It's actually a simplification of the precompilation as in almost all cases it will suffice to provide a single binary to support all relevant BEAM versions.

Got it. Makes sense!

Issue #509, opened 2022-12-28
Check if Term is float


juanazam:
Hello!

First of all, wanted to say this project is amazing, I'm a newbie at Rust and I was still able to write a nif and use it in Elixir!

While most of what I was set to code is working, there is one more thing I couldn't crack.

The nif I'm working on receives an Elixir map as an argument. The map can have different kinds of values (maps, lists, booleans, ints, floats, etc).

While decoding the Map as a Rust Struct, I'm having issues while decoding numbers. Since the map is dynamic, I don't know beforehand if the value I'm going to decode is an int or a float, and I haven't been able to make this work for both ints and floats at the same time.

This is the code I have:

```
if term.is_number() {
   let number_scalar: f64 = term.decode::<f64>().unwrap();
   ...
}
...
```

In order to call the right decoder, I need to specify the type but given I don't know the type, I can't do that.
I checked rustler and there is no `term.is_float` or `term.is_int` which could help me out.

I imagine the possibility of doing something like:

```
if term.is_int() {
   let number_scalar: i32 = term.decode::<i32>().unwrap();
   ...
} else if term.is_float() {
  let number_scalar: f64 = term.decode::<f64>().unwrap();
}
```

Is this possible? if it's not, does anybody has any recommendations on how to handle this case?

filmor:
- You should not use `unwrap` in "production" code. It will not crash your code but only because Rustler explicitly handles `panic`s
- Instead, you can use `if let` or `match` (or `unwrap_or`, there are plenty of options) to first try to decode as integer and, if that fails, decode as float instead
- @all: Is there any particular reason, why `Term::get_type()` is implemented using the various `is_*` methods (which map to `enif_is_*`)? Erlang's built in `enif_term_type` can actually distinguish floats and integers, changing the behaviour of `get_type` in this way now would be a breaking change, though.

juanazam:
@filmor Thanks so much for your reply! will implement this update!

evnu:
> Is there any particular reason, why Term::get_type() is implemented using the various is_* methods 

[enif_term_type](https://www.erlang.org/doc/man/erl_nif.html#enif_term_type) was introduced with OTP 22.0, so it was  not around when `get_type()` was implemented.

filmor:
Then I'll add it. @rusterlium/core do you think it's fine to break API here by adjusting the enum or should I add a new function?

evnu:
[...]

PR #538, opened 2023-05-14
Term Type

filmor:
Raise default version to 2.15 and implement get_type using enif_term_type

#509

### DIFF
--- a/rustler/src/dynamic.rs
+++ b/rustler/src/dynamic.rs
@@ -1,3 +1,6 @@
+#[cfg(feature = "nif_version_2_15")]
+use rustler_sys::ErlNifTermType;
+
 use crate::wrapper::check;
 use crate::Term;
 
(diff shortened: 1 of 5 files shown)


filmor:
The main disadvantage of this approach is that this actually breaks the API since `enif_term_type` distinguishes `float` and `integer`, but doesn't distinguish `list` and `empty_list` anymore.

evnu:
Seems that I forgot to save this comment:

Do we actually need to support nif versions below 2.15? Our README states that we support the three most-recent OTP releases, so I think that we can expect users to have at least NIF version 2.15. Note that we have an inconsistency in the README however:

> The minimal supported NIF version for a library should be defined via Cargo features. The default is currently 2.14 (Erlang/OTP 21). To use features from NIF version 2.16 (Erlang/OTP 24), the respective feature flag has to be enabled on the dependency:

filmor:
Good point, I'll adjust the README for this branch.

Let's release this (after the other pending PRs are processed) and drop 2.14 support after that.

PR #539, opened 2023-05-18
Add OTP26 to CI and bump rustler_sys

filmor:
- Bump rustler_sys patch version
- Add OTP26 to CI


filmor:
The errors stem from the now undefined term order in OTP26. I'll adjust the tests to "evaluate" the debug output s.t. we can compare on the term level.

PR #549, opened 2023-06-27
Only keep files as external resources

josevalim:


filmor:
There is already a very similar PR for this :) #548

josevalim:
perfect!

PR #550, opened 2023-06-27
Use compile_env for compile-time configuration

josevalim:
We also deprecate rustler_crates since that cannot be as efficiently tracked by the Elixir compiler.

### DIFF
--- a/rustler_mix/lib/rustler/compiler/config.ex
+++ b/rustler_mix/lib/rustler/compiler/config.ex
@@ -32,13 +32,18 @@ defmodule Rustler.Compiler.Config do
 
   alias Rustler.Compiler.Config
 
-  def from(otp_app, module, opts) do
-    config = Application.get_env(otp_app, module, [])
-
+  def from(otp_app, module, config, opts) do
     crate = config[:crate] || opts[:crate] || otp_app
 
     # TODO: Remove in 1.0
-    rustler_crates = Mix.Project.config()[:rustler_crates] || []
+    rustler_crates =
+      if mix_config = Mix.Project.config()[:rustler_crates] do
+        mix_config
+      else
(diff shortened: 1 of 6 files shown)


rustler_mix/lib/rustler/compiler/messages.ex
```diff
@@ -78,12 +62,6 @@ defmodule Rustler.Compiler.Messages do
     """
   end
 
-  def message({:cargo_no_name, crate}) do
-    """
-    No library or binary with name listed in Cargo.toml of crate '#{crate}'.
-    """
-  end
```
josevalim, line 85: Those messages are not raised from anywhere.

filmor:
Could you please update the "UPDATE.md" and changelog as well?

josevalim:
Done.

filmor:
I am preparing a point-release to get the Elixir 1.15 changes out before my vacation, I take it this one would not need to be included, right?

josevalim:
@filmor no need, yeah. :)

rustler_mix/lib/rustler/compiler/config.ex
```diff
@@ -32,13 +32,21 @@ defmodule Rustler.Compiler.Config do
...
+    rustler_crates =
+      if mix_config = Mix.Project.config()[:rustler_crates] do
+        mix_config
+      else
+        IO.warn(
+          ":rustler_crates in mix.exs is deprecated, please explicitly pass options on `use Rustler` or configure the module in your `config/*.exs` files"
+        )
+
+        []
+      end
```
filmor, line 48: @josevalim Am I misunderstanding something or is this logic the wrong way round? We should raise the warning if `:rustler_crates` is found, right?

PR #558, opened 2023-08-17
Extend load_from to allow loading from an absolute path

### LINKED ISSUE #556: Support for fully custom `:load_from` for escripts
Hi, I'm trying to use a single rust function inside my elixir app. I followed the setup instructions according to `mix rustler.new` and successfully got it working under `iex`. Unfortunately, I was getting an error when doing the same for `escript.build`, which is when I found out that rustler uses priv directory which is not accessible to escripts. Luckily, I came across the `load_from` configuration option and tried it in my module like this:

`use Rustler, otp_app: :fizzbuzz, crate: "fizzbuzz", load_from: {:fizzbuzz, "/Users/maheep/Downloads/native/libfizzbuzz"}`

Unfortunately, all it does is append the path to Application.app_dir output. Here's the problematic line in `rustler.ex`:

```elixir
load_path =
          otp_app
          |> Application.app_dir(path)
          |> to_charlist()
```

which I changed to:
```elixir
load_path =
          path
          |> to_charlist()
```

Doing this change allows me to specify fully custom `load_from` paths and gets things working for me.

MY QUESTION: Is `load_from` the correct option to use in my case? If so, do I have any other alternative (I don't think its a good idea to mess with a dependency's code).

MY OS: macOS 13.4.1 Ventura

evnu:
`load_from` already has a format to specify a path relative to an OTP application. This commit extends the type for `load_from` to also allow setting an absolute path. Note that no special check for the validity of the path is being done, as we do not know beforehand what the target operating system looks like. Setting this configuration correctly is thus the user's responsibility.

Fix #556.

### DIFF
--- a/rustler_mix/lib/rustler.ex
+++ b/rustler_mix/lib/rustler.ex
@@ -65,9 +65,18 @@ defmodule Rustler do
       loaded from at runtime. By default the compiled artifact is loaded from the
       owning `:otp_app`'s `priv/native` directory. This option comes in handy in
       combination with the `:skip_compilation?` option in order to load pre-compiled
-      artifacts. To override the default behaviour specify a tuple:
-      `{:my_app, "priv/native/<artifact>"}`. Due to the way `:erlang.load_nif/2`
-      works, the artifact should not include the file extension (i.e. `.so`, `.dll`).
+      artifacts. The configuration format:
+
+      ```
+      {otp_app :: atom(), path :: String.t()} | abspath :: String.t()}
+      ```
+
+      The tuple version defines that the artifact shall be found relative to `otp_app`,
+      for example `{:my_app, "priv/native/<artifact>"}`. The non-tuple alternative requires
+      an absolute path.
+
+      Due to the way `:erlang.load_nif/2` works, the artifact should not
+      include the file extension (i.e. `.so`, `.dll`).
 
     * `:mode` - Specify which mode to compile the crate with (default: `:release`)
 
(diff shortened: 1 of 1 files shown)


filmor:
Hmm, actually users should be able to just implement their own custom `on_load` handler right? I don't think it's a good idea to offer this in the library.

evnu:
>  I don't think it's a good idea to offer this in the library.

You mean offering `load_from` at all, or the extension proposed here?

filmor:
Passing the OTP application is fine, passing arbitrary paths is what I have an issue with. Nothing keeps users from just building their own `on_load` handler to achieve the same thing.

evnu:
@filmor to check that I get you right, you propose that users overwrite `@on_load` and avoid using the `rustler_init/0` function? That's prone to error as well I think. Erlang's `:erlang.load_nif/2` function does not restrict the path where the NIF library resides, I am not sure if we should impose restriction on that path "artificially".

filmor:
Yes... But that is exactly the problem with this PR as well. I suggest that if people want to load a NIF from an absolute path, they do it without `use Rustler`.

evnu:
> they do it without `use Rustler`.

That does not seem pratical. Users then also have to replicate the mechanism to auto-compile on changes with external resources and so on. I don't have a strong opinion on this topic, but I think users do not want to implement all of `use Rustler` just to load the library from a different path. But I see your point that using absolute paths introduces other problems for users. For example, the library is not necessarily versioned together with the application itself.

I will close this PR for now, we can discuss this further in #556 itself.

PR #559, opened 2023-08-17
Document to use runtime: false for rustler dependency

### LINKED ISSUE #516: Store rustler template outside of priv/?
The templates are stored in `rustler_mix/priv/templates`. When `rustler_mix` is part of a deployed Erlang/Elixir release, the priv directory is usually copied as-is into the release, so the templates end up in a deployed location. Maybe we can move the templates into another directory and tell `package/0` in `Rustler.Mixfile` to include that directory as well.

evnu:
As noted by @hauleth in #516, the rustler dependency is only required at compile time. This PR makes that possible by expanding `Rustler.construct_load_data/2` into the target module. The PR changes the documentation (and a test of the toolchain) to instruct that `runtime: false` can be used with rustler as a dependency.

Fix #516

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ See [`UPGRADE.md`](./UPGRADE.md) for additional help when upgrading to newer ver
 * Dropped support for `RUSTLER_NIF_VERSION`
 * Deprecated `:rustler_crates` project configuration
 * Mark `use Rustler` module configuration as compile-time
+* Make `:rustler` a compile-time-only dependency (#516, #559)
 
 ## [0.29.1] - 2023-06-30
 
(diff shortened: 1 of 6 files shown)


filmor (commented):
Could you update the Changelog and docs accordingly? I'm very much in favour of this, it looks like an oversight.

evnu:
I added a changelog entry and checked via `rg` that we refer to `runtime: false` everywhere:

```
~/tools/rustler (516-set-runtime_false ✔) rg ':rustler,'
rustler_tests/mix.exs
20:    [{:rustler, path: "../rustler_mix", runtime: false}]

prepare_release.sh
44:sed -i "s/{:rustler, \".*\"}/{:rustler, \"~> $VERSION\"}/" rustler_mix/README.md

rustler_mix/README.md
18:    {:rustler, "~> 0.29.1", runtime: false}

rustler_benchmarks/mix.exs
25:      {:rustler, path: "../rustler_mix", runtime: false}

rustler_mix/test.sh
50:  defp deps, do: [ {:rustler, path: "$rustler_mix", runtime: false} ]

rustler_mix/mix.exs
9:      app: :rustler,
```

filmor:
The prepare release script is not correct anymore, you'll need to drop the closing } for the regex to still match.

PR #561, opened 2023-08-28
Fix clippy lints

filmor:


filmor:
@evnu fyi

Cargo.toml
```diff
@@ -1,4 +1,5 @@
 [workspace]
+resolver = "2"
```
evnu, line 2: Should we split that and the change for edition 2021 into a separate PR? Or is it relevant to fixing the lints?

evnu (approved):
LGTM; Don't know if we should have an extra PR for the edition change to make it more explicit and add it to the changelog.

Cargo.toml
filmor: It fixes another warning, not a "fatal" one, though. I'll move it to a separate PR (and merge it directly).

PR #563, opened 2023-09-08
feat: add result-returning `send` variants to `Env` and `OwnedEnv`

bcksl:
`Env::send()` and `OwnedEnv::send_and_clear()` should ideally not elide the return value of [`enif_send`](https://www.erlang.org/doc/man/erl_nif#enif_send).

Included are several variants, all of which provide exactly the same information, given the current NIF API.

My take is that the preferred solution would be to pick one and resignature `send()`, but we could also keep some/all of the variants provided instead.

In all cases, resignature would not a breaking API change, but will generate warnings for the `_result` variants. The `_i32` and `_bool` variants are less idiomatic, but do not suffer from this issue. Nonetheless, my thinking is that an unhandled `Result` warning is ideal in this case, as it makes users aware of the fact that they are currently discarding information without breaking the build.

The naming is intentionally clunky, and if we decide to keep one or more of the `send_*()` variants rather than resignature `send()`, I'm happy to redraft with nicer names. Additional/modified tests will also follow that decision.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ See [`UPGRADE.md`](./UPGRADE.md) for additional help when upgrading to newer ver
 * Mark `use Rustler` module configuration as compile-time
 * Bump Rust edition to 2021
 * Make `:rustler` a compile-time-only dependency (#516, #559)
+* Return `Result<(), SendError>` from all `send` functions (#239, #563)
 
 ## [0.29.1] - 2023-06-30
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -8,6 +8,16 @@ This document is intended to simplify upgrading to newer versions by extending t
 options on `use Rustler` or configuring the module in your `config/*.exs`
 files.
 
+Additionally, `Env::send` and `OwnedEnv::send_and_clear` will now return a
+`Result`. Updating will thus introduce warnings about unused `Result`s. To
+remove the warnings without changing behaviour, the `Result`s can be "used" as
+```rust
+let _ = env.send(...)
+```
+Neither the `Ok` nor the `Err` case carry additional information so far. An
+error is returned if either the receiving or the sending process is dead. See
+also [enif\_send](https://www.erlang.org/doc/man/erl_nif.html#enif_send).
+
 ## 0.28 -> 0.29
 
 `RUSTLER_NIF_VERSION` is deprecated and will not be considered anymore for 0.30.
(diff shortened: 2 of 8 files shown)


filmor:
I think only the `bool` and `Result<(), ()>` variants make sense. Alternatively, you could add a (for now empty) error type `pub struct SendError;`.

I'm not quite sure, where you would see an improvement with this, though. The two documented error states (sender process is not alive or receiver process is not alive) are only checked at the time of sending, it could still be that the receiver process dies before processing the message. I'm also not sure whether the first case (sender process died) is even possible with our interface, at least not for owned envs. It is definitely wrong to just blindly `unwrap` the result here. Additionally, Erlang's `!` operator also doesn't report any of this information back.

filmor:
Thinking a bit more about this, I'd propose to change the signature to `send(...) -> bool`. As said, I don't think we should encourage people to aggressively check the result if they don't know what they are doing, certainly not with an easy way to cause a panic from it. Exposing it makes sense, though, and conceptually it is a boolean, after all.

@evnu @hansihe @scrogson Opinions?

evnu:
I think whether to use a `bool` or `Result` comes down to how a user wants to handle failure in this case, and if the extra information with `Result` would be useful. With `Result`, we can distinguish if the receiver failed (for the case of `OwnedEnv`), or if either the sender or the receiver failed. Now, if this is actually useful to handle the error is another question. I'd expect that usually a failure with this function would only be logged and maybe retried. If that is so, a bool is probably just fine.

> As said, I don't think we should encourage people to aggressively check the result if they don't know what they are doing, certainly not with an easy way to cause a panic from it.

But maybe we should encourage users to read the documentation for that function, as the boolean fuses two distinct error cases. That is useful information when debugging why an expected message might not turn up at a receiver. When using a `Result` which needs to be handled, somebody who reads the code using `send()` will also have an easier time seeing that this might error out, even if that reader does not know the documentation (yet).

In any case, I don't have a strong opinion on this one. I can see benefits in both. As `-> bool` simply forwards what the NIF API exposes, we should probably go with that and avoid documentation friction between the rustler documentation and what the NIF API defines.

filmor:
I have a strong preference for `fn send(...) -> bool`. So if no one else has a strong opinion, I'd say we go with that :)

hansihe:
An advantage of using a Result might be that the user is actually warned if they ignore the return value.

It forces the user to deal with the possibility of error, even if they just end up doing `let _ = [send]`

filmor:
I have incorporated the results of this discussion (going with returning `Result<(), SendError>` now) and added tests for the simpler case of the receiver being dead. I'm not quite sure how to test the other case (sender dead), but it seems to be quite esoteric anyhow.

evnu (approved):
LGTM.

PR #568, opened 2023-10-14
POC PGO configuration

cleaton:
This PR adds a POC implementation to enable rustler projects to configure scripts that will be used to optimize the rust code using PGO following: https://doc.rust-lang.org/rustc/profile-guided-optimization.html#a-complete-cargo-workflow

I've only tested it on the Struct benchmark, it seems to give small but consistent (I only tried running a few times manually) improvement especially for the bigger "decode & encode" benchmark:
```
Benchmarking decode ...
Benchmarking decode and encode ...
# With PGO
Name                        ips        average  deviation         median         99th %
decode                   1.06 M        0.94 μs    ±49.74%        0.92 μs        1.30 μs
decode and encode        0.80 M        1.26 μs  ±1450.35%        1.20 μs        1.46 μs

Comparison: 
decode                   1.06 M
decode and encode        0.80 M - 1.33x slower +0.31 μs


# Without PGO
Name                        ips        average  deviation         median         99th %
decode                   1.04 M        0.96 μs    ±18.04%        0.95 μs        1.28 μs
decode and encode        0.76 M        1.32 μs  ±1367.14%        1.23 μs        2.17 μs

Comparison: 
decode                   1.04 M
decode and encode        0.76 M - 1.38x slower +0.36 μs
```

Maybe there is a bigger rustler project we could try it on.

This was first discussed in #565

### DIFF
--- a/rustler_mix/lib/rustler/compiler.ex
+++ b/rustler_mix/lib/rustler/compiler.ex
@@ -6,44 +6,90 @@ defmodule Rustler.Compiler do
   @doc false
   def compile_crate(otp_app, config, opts) do
     config = Config.from(otp_app, config, opts)
+    pgo_run = System.get_env("RUSTLER_PGO_RUN") === "true"
 
-    unless config.skip_compilation? do
-      crate_full_path = Path.expand(config.path, File.cwd!())
-
-      File.mkdir_p!(priv_dir())
-
-      Mix.shell().info("Compiling crate #{config.crate} in #{config.mode} mode (#{config.path})")
-
-      [cmd | args] =
-        make_base_command(config.cargo)
-        |> make_no_default_features_flag(config.default_features)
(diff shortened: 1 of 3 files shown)


rustler_mix/lib/rustler/compiler.ex
```diff
@@ -6,45 +6,91 @@ defmodule Rustler.Compiler do
...
+    # in `_build`.
+    Mix.Project.build_structure()
     config
   end
 
+  defp do_pgo(%Config{pgo_commands: cmds} = config) when is_list(cmds) do
+    Mix.shell().info("Configuring build for PGO profile collection")
+
+    # TODO better folder
+    File.rmdir("/tmp/pgo-data")
```
evnu, line 60: Should probably be somewhere in the build path to make that compatible with windows.

PR #573, opened 2023-10-26
Fix warnings for non-snake case atom module names

bcksl:


turion:
LGTM. I think these are exactly the cases where such warnings are emitted.

turion (approved):

PR #574, opened 2023-11-09
Add support for generic types

### LINKED ISSUE #424: NifTuple doesn't work with polymorphic structs

turion:
Closes #424. Allows generic type parameters in the derive macros:

```rust
    #[derive(NifStruct)]
    #[module = "GenericStruct"]
    pub struct GenericStruct<T> {
        t: T
    }
```

### DIFF
--- a/rustler_codegen/src/encode_decode_templates.rs
+++ b/rustler_codegen/src/encode_decode_templates.rs
@@ -31,25 +31,25 @@ pub(crate) fn decoder(ctx: &Context, inner: TokenStream) -> TokenStream {
         let where_clause = impl_generics.make_where_clause();
 
         for lifetime in lifetimes {
-            let mut puncated = syn::punctuated::Punctuated::new();
-            puncated.push(lifetime.clone());
+            let mut punctuated = syn::punctuated::Punctuated::new();
+            punctuated.push(lifetime.clone());
             let predicate = syn::PredicateLifetime {
                 lifetime: decode_lifetime.clone(),
                 colon_token: syn::token::Colon {
                     spans: [Span::call_site()],
                 },
-                bounds: puncated,
+                bounds: punctuated,
             };
             where_clause.predicates.push(predicate.into());
 
-            let mut puncated = syn::punctuated::Punctuated::new();
-            puncated.push(decode_lifetime.clone());
+            let mut punctuated = syn::punctuated::Punctuated::new();
+            punctuated.push(decode_lifetime.clone());
             let predicate = syn::PredicateLifetime {
                 lifetime: lifetime.clone(),
                 colon_token: syn::token::Colon {
                     spans: [Span::call_site()],
                 },
-                bounds: puncated,
+                bounds: punctuated,
             };
             where_clause.predicates.push(predicate.into());
         }
--- a/rustler_codegen/src/encode_decode_templates.rs
+++ b/rustler_codegen/src/encode_decode_templates.rs
@@ -1,5 +1,6 @@
 use proc_macro2::{Span, TokenStream};
 use quote::quote;
+use syn::{GenericArgument, PathSegment, TraitBound};
 
 use super::context::Context;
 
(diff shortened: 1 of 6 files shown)


turion:
CC @evnu this obsoletes #425

turion:
I'm not sure where I should add documentation.

rustler_codegen/src/encode_decode_templates.rs
```diff
@@ -31,25 +32,88 @@ pub(crate) fn decoder(ctx: &Context, inner: TokenStream) -> TokenStream {
...
+        let where_clause = impl_generics.make_where_clause();
+
+        for type_parameter in type_parameters {
+            let mut punctuated = syn::punctuated::Punctuated::new();
+            punctuated.push(decode_lifetime.clone().into());
+            punctuated.push(syn::TypeParamBound::Trait(TraitBound {
+                paren_token: None,
+                modifier: syn::TraitBoundModifier::None,
+                lifetimes: None,
+                // path: syn::Ident::new("Decoder", Span::call_site()).into(),
```
evnu, line 78:
```suggestion
```

rustler_codegen/src/encode_decode_templates.rs
```diff
@@ -31,25 +32,88 @@ pub(crate) fn decoder(ctx: &Context, inner: TokenStream) -> TokenStream {
...
+                        },
+                    ]
+                    .iter()
+                    .cloned()
+                    .collect(),
+                },
+            }));
+            let predicate = syn::PredicateType {
+                lifetimes: None,
+                // bounded_ty: syn::Type::Verbatim(type_parameter.ident.to_token_stream()),
```
evnu, line 108:
```suggestion
```

rustler_codegen/src/encode_decode_templates.rs
```diff
@@ -69,7 +133,56 @@ pub(crate) fn decoder(ctx: &Context, inner: TokenStream) -> TokenStream {
...
+    let type_parameters: Vec<_> = generics
+        .params
+        .iter()
+        .filter_map(|g| match g {
+            syn::GenericParam::Type(t) => Some(t.clone()),
+            _ => None,
+        })
+        .collect();
+
+    if !type_parameters.is_empty() {
```
evnu, line 146: I think we can remove this `if` and iterate over an empty `type_parameters` instead. With that, we reduce indentation, which makes this more readable.

rustler_codegen/src/encode_decode_templates.rs
```diff
@@ -31,25 +32,88 @@ pub(crate) fn decoder(ctx: &Context, inner: TokenStream) -> TokenStream {
...
```
evnu, line 68: Do we need this `if`, or could we instead just have the loop running over an empty `type_parameters`?

rustler_codegen/src/encode_decode_templates.rs
```diff
@@ -69,7 +133,56 @@ pub(crate) fn decoder(ctx: &Context, inner: TokenStream) -> TokenStream {
 
 pub(crate) fn encoder(ctx: &Context, inner: TokenStream) -> TokenStream {
     let ident = ctx.ident;
-    let generics = ctx.generics;
+    let mut generics = ctx.generics.clone();
+    let type_parameters: Vec<_> = generics
```
evnu, line 137: This is the same filter as further above, we should extract that into a function to avoid duplication.

rustler_codegen/src/encode_decode_templates.rs
```diff
@@ -31,25 +32,88 @@ pub(crate) fn decoder(ctx: &Context, inner: TokenStream) -> TokenStream {
...
                     spans: [Span::call_site()],
                 },
-                bounds: puncated,
+                bounds: punctuated,
+            };
+            where_clause.predicates.push(predicate.into());
+        }
+    }
+
+    let type_parameters: Vec<_> = generics
```
evnu, line 59: What kind of generics do we need to skip here? That's not clear from this source location alone. A comment might help.

evnu (changes requested):
Thanks, good work! Some comments from my side.

rustler_codegen/src/encode_decode_templates.rs
turion: True, I copied the pattern from the lifetimes clause before.

rustler_codegen/src/encode_decode_templates.rs
turion: Into  a function or rather a field of `Context`?

evnu: I guess a field is better :)

rustler_codegen/src/encode_decode_templates.rs
turion: I would remove the corresponding `if` from the lifetimes as well, because if I reduce the indentation I need to unconditionally create a `where` clause.

turion: The only other way would be calling `generics.make_where_clause()` inside the loop. I believe this would work because it's idempotent, but I don't know what's the best way.

rustler_codegen/src/encode_decode_templates.rs
turion: I'd refactor the lifetimes as well in this case.

rustler_codegen/src/encode_decode_templates.rs
turion: I added a comment, does it help?

evnu: Yep, thanks.

rustler_codegen/src/encode_decode_templates.rs
evnu: Removing the corresponding if makes sense I guess, the parts should look fairly similar in the end.

evnu (approved):
LGTM.

Issue #577, opened 2023-11-20
Can't `mix compile` with path dependency in `Cargo.toml`


grzuy:
```
== Compilation error in file lib/candlex/native.ex ==
** (Protocol.UndefinedError) protocol Enumerable not implemented for nil of type Atom
    (elixir 1.15.7) lib/enum.ex:1: Enumerable.impl_for!/1
    (elixir 1.15.7) lib/enum.ex:166: Enumerable.reduce/3
    (elixir 1.15.7) lib/enum.ex:4387: Enum.filter/2
    (rustler 0.30.0) lib/rustler/compiler/config.ex:130: Rustler.Compiler.Config.gather_local_crates/4
    (rustler 0.30.0) lib/rustler/compiler/config.ex:108: Rustler.Compiler.Config.external_resources/3
    (rustler 0.30.0) lib/rustler/compiler/config.ex:79: Rustler.Compiler.Config.build/1
    (rustler 0.30.0) lib/rustler/compiler.ex:8: Rustler.Compiler.compile_crate/3
    lib/candlex/native.ex:11: (module)
```

grzuy:
Concretely happening for me in https://github.com/mimiquate/candlex/compare/candle-core-path?expand=1.

grzuy:
In this case the crate dependency is a workspace member and the one failing is another workspace member of the same workspace.
Might be an edge case causing this?

PR #578, opened 2023-11-20
fixes Cargo.toml local path dep?

grzuy:
A (naive?) workaround for #577.
Opening as draft at least to show where I landed when troubleshooting the issue.


filmor:
This is the same as #607 which is merged now.

grzuy:
Thank you!

PR #579, opened 2023-11-20
Fix error messages

turion:
While hacking on the codebase we noticed that the error messages when creating an incorrect `tag` or `module` are wrong.


turion:
The pipeline is a bit flaky it seems, the macos runner had network trouble. But I don't have the privileges to rerun the CI.

PR #212, opened 2019-05-21
Add support for OTP-22

filmor:
Fixes #211.

PR #581, opened 2023-12-29
Fix Term's "get_type()" implementation

philss:
It was not returning the correct type for integers, probably due to a "garbage" arriving because of the wrong type.
This affects the latest v0.30.

### DIFF
--- a/rustler/src/dynamic.rs
+++ b/rustler/src/dynamic.rs
@@ -43,7 +43,7 @@ impl From<ErlNifTermType> for TermType {
 }
 
 pub fn get_type(term: Term) -> TermType {
-    if cfg!(nif_version_2_15) {
+    if cfg!(feature = "nif_version_2_15") {
         term.get_erl_type().into()
     } else if term.is_atom() {
         TermType::Atom
--- a/rustler/src/dynamic.rs
+++ b/rustler/src/dynamic.rs
@@ -1,3 +1,5 @@
+use std::ffi::c_double;
+
 #[cfg(feature = "nif_version_2_15")]
 use rustler_sys::ErlNifTermType;
 
@@ -56,7 +58,11 @@ pub fn get_type(term: Term) -> TermType {
     } else if term.is_map() {
         TermType::Map
     } else if term.is_number() {
-        TermType::Float
+        if term.is_float() {
+            TermType::Float
+        } else {
+            TermType::Integer
+        }
     } else if term.is_pid() {
         TermType::Pid
     } else if term.is_port() {
@@ -98,4 +104,15 @@ impl<'a> Term<'a> {
     impl_check!(is_port);
     impl_check!(is_ref);
     impl_check!(is_tuple);
+
+    pub fn is_float(self) -> bool {
+        let mut val: c_double = 0.0;
+        unsafe {
+            rustler_sys::enif_get_double(self.get_env().as_c_arg(), self.as_c_arg(), &mut val) == 1
+        }
+    }
+
+    pub fn is_integer(self) -> bool {
+        self.is_number() && !self.is_float()
+    }
 }
--- a/rustler/src/dynamic.rs
+++ b/rustler/src/dynamic.rs
@@ -43,7 +43,7 @@ impl From<ErlNifTermType> for TermType {
 }
 
 pub fn get_type(term: Term) -> TermType {
-    if cfg!(feature = "nif_version_2_15") {
+    if cfg!(feature = "nif_version_2_15") && !cfg!(target_family = "windows") {
         term.get_erl_type().into()
     } else if term.is_atom() {
         TermType::Atom
--- a/rustler_sys/build.rs
+++ b/rustler_sys/build.rs
@@ -848,7 +848,7 @@ fn build_api(b: &mut dyn ApiBuilder, opts: &GenerateOptions) {
         b.func(
             "ErlNifTermType",
             "enif_term_type",
-            "env: *mut ErlNifEnv, term: *const ERL_NIF_TERM",
+            "env: *mut ErlNifEnv, term: ERL_NIF_TERM",
         );
 
         b.func("c_int", "enif_is_pid_undefined", "pid: *const ErlNifPid");
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -19,8 +19,9 @@ This document is intended to simplify upgrading to newer versions by extending t
    See also
    [enif\_send](https://www.erlang.org/doc/man/erl_nif.html#enif_send).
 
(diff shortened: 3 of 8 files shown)


philss:
Looks like this is not working well on Windows. I'm going to investigate tomorrow.

rustler/src/dynamic.rs
filmor, line 1: This doesn't fix anything, right? In that case, please revert.

rustler_sys/build.rs
filmor, line 1: Nice catch, this was already wrong in the old Erlang-based generation!

filmor (approved):

filmor:
Hah, even found out who introduced the bug: https://github.com/rusterlium/rustler/pull/212 🕵️‍♂️

I can't debug the Windows issue for the next week or so since I won't have a Windows machine available.

rustler/src/dynamic.rs
philss: It was fixing, but it could keep the same style: https://github.com/rusterlium/rustler/pull/581/commits/d61e4bc7dfd9205d0ce0d609d64c0c3086601a0a

rustler_sys/build.rs
philss: Yeah, I think we didn't realize because it was not in use directly.

philss:
> I can't debug the Windows issue for the next week or so since I won't have a Windows machine available.

Thanks! I'm trying to setup a Windows VM so I can debug this as well :)

philss:
Just to give an update: I've tested and was able to reproduce the build problem on Windows.
Unfortunately I couldn't yet figure out what is going on - it just return "unknown" for any term there :/
I should go back to this next week.

filmor:
Yeah, I'll be able to look into it this weekend as well. My initial guess would be a mismatched base type for the enum, Windows `long`s are always 32 bit.

filmor:
@philss Did you get any further? I couldn't reproduce it so far, but I'll try again tomorrow.

philss:
@filmor unfortunately I couldn't :/

I tried to figure out where the int type is wrong, but there is no explicit declaration of it.
For a comparison, I did write a small NIF in C and it works good on Windows. So the problem seems to be on our side.
I will give another try this week as well.

philss:
@filmor I tried a bunch of things to fix this problem with the Windows build, but I couldn't find a solution.
I suspect the problem is in the bindings we create in the `rustler_sys` create for the NIF API.

I also tried to setup [`rust-bindgen`](https://github.com/rust-lang/rust-bindgen) on my windows VM, but without success. The idea was to generate a file that would follow more closely what is expected from the `erl_nif.h` file.

I think, given that we cannot make it work on Windows easily, we could drop this feature for that OS, documenting this behaviour. WDYT?

josevalim:
Given it never worked, perhaps we should just nuke it altogether. :)

filmor:
Well, it works fine on Unix and is IMO the preferable approach (single function call vs. potentially ~10). I'll try to get the Windows interface in line (I think this is mostly about distinguishing floats from integers).

josevalim:
It is a single function call but internally it is 10 bitmap comparisons. If you want to know if something is a float or integer, asking specifically for float and integer would be more efficient.

Also note that nor Erlang nor Elixir have a similar function and we have been just fine for decades. Given it is broken in main, odds are that we really don’t need it.

that said, if you can make it work, then that’s great. I am just saying that, if it doesn’t work, it is not the end of the world. :)

filmor:
@philss @evnu This finally works and passes all tests. It's really not pretty to essentially "copy" this gnarly code for `is_float`. We could still do as Jose suggests and drop the whole new code path, but since it works fine on non-Windows (which is likely the primary userbase anyhow) and we now have a workaround.

### Notes:

- I will still have to test this manually on 32bit Windows as that is not part of our CI at all and the `is_float` codepath is totally different (maybe `enif_term_type` works there as well?)
- There is apparently a bug in OTP24, in that `is_fun` doesn't work for remote funs (`fun module:function/0` or in Elixir `&Function.identity/1`). I don't intend to debug this further, it works for "function objects".

josevalim:
For what is worth, I don't think we should export a function that does not work on Windows as it may become an additional pitfall for both developers and Windows users. When a package using Rustler misbehaves on Windows, it may lead to both users and authors spent time debugging only to land here.

If you really want to go down that route, can it be an opt-in feature? So it is not available by default, reducing the chance of accidental use. Or, make sure that it is not available on Windows (so it is not even compiled) or that it raises a clear error message?

filmor:
Which export are you now talking about? `enif_term_type` is in `rustler-sys`, which is like all`-sys` crates "there be dragons". What exactly would you make optional here? My intention is to eventually get rid of the custom `is_float` again and use `enif_term_type` everywhere. The way it's now implemented, everything works the same way everywhere (apart from checking remote funs on OTP24, but that is a bug in a version of OTP that we are going to stop supporting in a few months (whem 27 comes out) anyhow.

josevalim:
My comment was to make `enif_term_type` optional. :) However, I was assuming that it still does not work on Windows (while the C NIF does). Given you said:

> The way it's now implemented, everything works the same way everywhere, apart from checking remote funs on OTP24

If it all works on Windows too, then just ignore me. :) But it may be that using `enif_get_float` and attempting to get a float may be more efficient than the several masks from `enif_term_type`.

filmor:
- On all non-Windows systems with OTP>=22 we use `enif_term_type` for the `get_type` function
- On Windows, we use the old way with an additional hacky-copied version of the non-exposed `is_float` (which I will probably replace with an `enif_get_float`)
- The behavior is now identical, the implementation is not (yet)
- The bug with remote funs on OTP24 is preexisting ant not something I intend to follow up on apart from writing it down
- I think the `get_type` will be useful for serializations, if benchmarks show that that's slower than chaining the `enif_is_*`, I'd consider that a bug in OTP since that is accoding to the docs why `enif_term_type` was introduced

philss (commented):
Awesome! For me this is good enough for now 🚢 

@filmor thank you for the efforts here!

rustler/src/dynamic.rs
```diff
@@ -98,4 +102,46 @@ impl<'a> Term<'a> {
...
+            1
+        } else {
+            2
+        };
+
+        const HEADER_FLONUM: usize = (SIZE << HEADER_ARITY_OFFSET) + TAG_HEADER_FLOAT;
+
+        let term_val = self.as_c_arg();
+        if term_val & TAG_PRIMARY_MASK == TAG_PRIMARY_BOXED {
+            dbg!(term_val);
```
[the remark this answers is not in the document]
filmor: Thanks, I'll remove this in one go when replacing this with `enif_get_double`

PR #591, opened 2024-02-13
Add Elixir 1.16 to pipeline

filmor:


.github/workflows/main.yml
```diff
@@ -85,7 +85,8 @@ jobs:
     strategy:
       matrix:
         pair:
-          - { erlang: "26", elixir: "1.15", latest: true }
+          - { erlang: "26", elixir: "1.16", latest: true }
+          - { erlang: "26", elixir: "1.15" }
           - { erlang: "26", elixir: "1.14" }
           - { erlang: "25", elixir: "1.14" }
           - { erlang: "24", elixir: "1.13" }
```
evnu, line 92: Should we drop `1.13` with that? At least our current policy says we support the three newest OTP versions and the three newest Elixir versions.

filmor: I just didn't want to bother checking whether 1.14 supports OTP24 (which we still have to check). But yeah, we should probably adjust it.

evnu: According to [the docs](https://hexdocs.pm/elixir/1.16.1/compatibility-and-deprecations.html), Elixir 1.14 supports OTP 23-25 (and OTP 26 as of 1.14.5):

image

filmor: How about 24+1.14, 25+1.15, 26+1.15 and 26+1.16?

PR #593, opened 2024-02-15
Allow arbitrary function names

filmor:
Allow arbitrary ASCII strings as NIF names. Follow-up of #505.

### DIFF
--- a/rustler_codegen/src/nif.rs
+++ b/rustler_codegen/src/nif.rs
@@ -48,15 +48,19 @@ pub fn transcoder_decorator(nif_attributes: NifAttributes, fun: syn::ItemFn) -> 
     let argument_names = create_function_params(inputs.clone());
     let erl_func_name = nif_attributes
         .custom_name
-        .map(|n| syn::Ident::new(n.value().as_str(), Span::call_site()))
-        .unwrap_or_else(|| name.clone());
+        .map(|n| n.value().to_string())
+        .unwrap_or_else(|| name.clone().to_string());
+
+    if !erl_func_name.is_ascii() {
+        panic!("Only ASCII strings are supported as function names");
+    }
 
     quote! {
         #[allow(non_camel_case_types)]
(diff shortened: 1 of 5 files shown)


rustler_codegen/src/nif.rs
```diff
@@ -48,15 +48,19 @@ pub fn transcoder_decorator(nif_attributes: NifAttributes, fun: syn::ItemFn) ->
     let argument_names = create_function_params(inputs.clone());
     let erl_func_name = nif_attributes
         .custom_name
-        .map(|n| syn::Ident::new(n.value().as_str(), Span::call_site()))
-        .unwrap_or_else(|| name.clone());
+        .map(|n| n.value().to_string())
+        .unwrap_or_else(|| name.clone().to_string());
+
+    if !erl_func_name.is_ascii() {
```
evnu, line 54: That makes me wonder if any control character could be used as a function name :)

rustler_codegen/src/nif.rs
```diff
@@ -48,15 +48,19 @@ pub fn transcoder_decorator(nif_attributes: NifAttributes, fun: syn::ItemFn) ->
     let argument_names = create_function_params(inputs.clone());
     let erl_func_name = nif_attributes
         .custom_name
-        .map(|n| syn::Ident::new(n.value().as_str(), Span::call_site()))
-        .unwrap_or_else(|| name.clone());
+        .map(|n| n.value().to_string())
```
evnu, line 51: I think this could be fused with the unwrap below using `map_or_else()` (which does both rewrap with `Option`), but this here also works just fine 🙂

evnu (approved):
LGTM, thanks! :) thanks @KoviRobi for the original PR!

rustler_codegen/src/nif.rs
filmor: Good point, we'll just reject those.

Issue #594, opened 2024-02-28
[Feature request] Add support for i128 and u128


Munksgaard:
Since both Rust and Elixir have support for 128-bit sized integers ([i128](https://doc.rust-lang.org/std/primitive.i128.html), [u128](https://doc.rust-lang.org/std/primitive.u128.html), [erlang](https://www.erlang.org/doc/efficiency_guide/advanced#memory)) it seems natural that Rustler should support it as well.

I have a specific use case for i128s as well, which is implementing support for the [Decimal](https://docs.rs/polars/latest/polars/datatypes/enum.AnyValue.html#variant.Decimal) type from Polars in [Explorer](https://github.com/elixir-explorer/explorer/issues/867).

Munksgaard:
I can't really figure out if [this](https://github.com/rust-lang/rust/issues/54341) is going to be a problem, but at least it's something to be aware of.

evnu:
Would this work for you: https://github.com/rusterlium/rustler/tree/master/rustler_bigint ?

filmor:
> I can't really figure out if [this](https://github.com/rust-lang/rust/issues/54341) is going to be a problem, but at least it's something to be aware of.

I don't think this is relevant for Rustler. It would be, if the NIF API actually included (C-based) support for 128-bit integer types, but it only supports up to 64.

If someone wants to work on this, https://github.com/rusterlium/rustler/tree/master/rustler_bigint can be taken as a basis. While that library is more general and implements support for arbitrarily-sized integers, the same approach (going via the `term_to_binary`) can be used for the standard library types `i128` and `u128`.

Munksgaard:
I have not had time to look at the bigint library yet, but it indeed seems like the easiest way to handle something like this.

Munksgaard:
Incredible! Thank you for your work @filmor!

PR #600, opened 2024-03-12
Implement Decoder/Encoder support for i/u128

filmor:
Implements #594.

### DIFF
--- a/rustler/src/types/i128.rs
+++ b/rustler/src/types/i128.rs
@@ -1,0 +1,130 @@
+use crate::{Decoder, Encoder, Env, Error, NifResult, Term};
+use std::convert::TryFrom;
+
+const EXTERNAL_TERM_FORMAT_VERSION: u8 = 131;
+const SMALL_BIG_EXT: u8 = 110;
+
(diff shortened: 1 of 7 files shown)


rustler/src/types/i128.rs
```diff
@@ -0,0 +1,130 @@
...
+            return Ok(int as i128);
+        }
+
+        let input = term.to_binary();
+        let input = input.as_slice();
+        if input.len() < 4 {
+            return Err(Error::BadArg);
+        }
+
+        if input[0] != EXTERNAL_TERM_FORMAT_VERSION {
```
evnu, line 61: Can we use `<` here to avoid breaking code when the ETF version is bumped?

filmor: The ETF version is never bumped amd all libraries hard code it (e.g. https://github.com/sile/eetf/blob/40d786a9e47d7c104466c94e69df8b03afcca5ca/src/codec.rs#L57).

evnu (approved):
LGTM!

PR #601, opened 2024-03-12
Replace rustler_bigint by a feature flag on rustler

filmor:


evnu (approved):
LGTM! :)

filmor:
We could publish a version with a deprecation warning, yes. I don't see any need to rush this, though. The library will keep working even with this change and the necessary steps are listed in the `UPDATE` docs.

PR #605, opened 2024-03-27
Update README.md

dvic:
Without compiling the dependencies I get

```
❯ mix help rustler.new
** (Mix) The task "rustler.new" could not be found
```


evnu:
Thanks!

Issue #598, opened 2024-03-07
[Feature request] Support for serializing Path/PathBuf


begleynk:
Would it be possible to implement [Encoder](https://docs.rs/rustler/latest/rustler/types/trait.Encoder.html)/[Decoder](https://docs.rs/rustler/latest/rustler/types/trait.Decoder.html) for Rust's [Path](https://doc.rust-lang.org/std/path/struct.Path.html)/[Pathbuf](https://doc.rust-lang.org/std/path/struct.PathBuf.html) types?

I'm currently working on a project where I own all the types in a complex struct, but because the standard library's Path/PathBuf can't be serialized over to Elixir, it's forcing me to create wrapper types. This would be a huge ergonomic benefit.

Happy to contribute a PR if this is straightforward.

Thanks for all your work on Rustler ❤️

filmor:
If all you are looking for is `Path(Buf) <-> binary string`, I don't see a problem with that.

begleynk:
Yup that's fine by me 👍 The Elixir [`Path`](https://hexdocs.pm/elixir/1.13/Path.html) module just works on binary strings, so this seems reasonable.

I'll try to contribute a PR for this in the coming days/weeks.

PR #608, opened 2024-05-02
Implement support for conversions of paths

filmor:
#598


evnu (approved):
LGTM. I wondered if a path could legally be non-utf8 and thus fail on decode, wdyt?

filmor:
It could be, but I think it's a valid assumption that binaries from the BEAM that are supposed to represent strings are encoded UTF-8. If we ever get a request from someone who wants this to work with different encodings, we can act at that time :)

PR #609, opened 2024-05-03
Drop lazy_static in favour of std::sync::OnceLock

filmor:


evnu (approved):
LGTM. I wonder if moving the MSRV to 1.70 is steep for users. No idea which older versions of Rust are commonly used.

PR #613, opened 2024-05-28
Discover NIFs at startup

filmor:
This is a proof-of-concept for a mechanism that does away with the need to list all NIF functions in the `rustler::init!` invocation. A similar mechanism will be required and implemented for resource type registrations.

- Convert `Nif` to a `struct`
- Use `inventory` to register the `Nif` instances
- In `rustler::init!`, ignore the passed functions and instead use the registered `Nif` instances

### Next steps:

- [x] See whether `linkme` is not a better option (creates the array directly at build-time, no need to leak) (/edit Looking at this again, don't think this adds much, the current implementation is fine)
- [x] Implement resource registration using this mechanism (#617)
- [ ] See if we can move the macros to `rustler` (might do this in another iteration, no need to put it in the same PR)


scrogson:
Love this idea!

rustler_codegen/src/init.rs
```diff
@@ -1,26 +1,28 @@
...
 impl Parse for InitMacroInput {
     fn parse(input: ParseStream) -> Result<Self> {
         let name = syn::Lit::parse(input)?;
-        let _comma = <syn::Token![,]>::parse(input)?;
-        let funcs = syn::ExprArray::parse(input)?;
+
+        if input.peek(syn::token::Comma) && input.peek2(syn::token::Bracket) {
+            let _ = syn::token::Comma::parse(input);
+            let _funcs = syn::ExprArray::parse(input);
+            // TODO: Generate deprecation warning
```
scrogson, line 19: do we want to add this now?

rustler_tests/native/rustler_test/src/test_resource.rs
```diff
@@ -109,7 +109,7 @@ pub fn resource_make_binaries(
     )
 }
 
-#[rustler::nif]
+// #[rustler::nif]
```
scrogson, line 112: Is there no test for this?

filmor: Nope. I'll try to trace back why. Maybe this was superceded by a different NIF. With the new logic (and this will need to go in the docs), one has to `cfg`-disable unused NIFs.

filmor: I see, I introduced it here: https://github.com/rusterlium/rustler/pull/487/files

Looks like I merged the to-be-tested behaviour into the single `resource_make_binaries` nif.

rustler_codegen/src/init.rs
filmor: Done

filmor:
From my point of view, this is done and ready to merge.

scrogson (approved):
:shipit:

filmor:
I'll merge this now. We can further discuss on #620 and/or #617 whether and how to use `linkme` instead.

PR #614, opened 2024-05-29
cargo-rustler

filmor:
Initial PoC for a tool that takes a compiled NIF library and reads back its information. This could be used to automatically generate a matching Erlang (`rebar3_cargo`) or Elixir (`Rustler`) module. Could be a dead end, but I don't like the current ergonomics of the wrapper modules.


hansihe:
An alternate approach here might be to serialize the configuration and embed it in the binary as a symbol. We could then just read it directly from the binary, without having to load it.

I think this would have a few advantages:
* Reading out the config could be fully platform independent 
* It would not require us to actually load the library to get the config
* We don't need a native tool, and we don't need to provide all symbols the NIFs require

filmor:
Any example of what you are thinking of? With the NIF discovery merged, the NIF collection will only be known when loading the library (which is how NIF linraries work on general).

Issue #84, opened 2017-02-17
Add ATO Resources


goertzenator:
I've been churning through many ideas on my Ruster project lately, and one that worked out really well and I think would be useful for Rustler is ["Any Trait Object Resources" (ATO Resources)](https://github.com/goertzenator/ruster/blob/binder/src/resource.rs)

In a nutshell, with a single resource type registration you can use any Rust type as a resource.  The advantages are:
- No need to pre-register types. You just start using them and they work.
- No resource traits so no need for a wrapper type to get around trait orphan rules.  For example `Cell<String>` can be used directly as a resource type.

So, if this fits in for Rustler I invite you to borrow the code; I will happily answer any questions about it.  I'll also point out that the upcoming `enif_select()` function in OTP-20 will likely force a revisit of resources.

jorendorff:
Cool. I considered something like this for Rustler, but didn't think @hansihe would like the slight extra allocation overhead or the extra runtime overhead of the type check. I couldn't figure out how to avoid either one.

(It seems like `Any` should have its own drop-in-place method—Rust "knows" in some sense how to drop an `Any` trait object, since `Box<Any>` works—but it doesn't.)

jorendorff:
Oh, it looks like your code avoids all but a word of allocation overhead. That seems close enough. It's just the runtime check then; and that will be a well-predicted branch.

hansihe:
This is a very cool idea, and it would improve usability a lot.

If we decide to add this, would it make sense to keep the existing way of doing it or should we get rid of it?

PR #617, opened 2024-06-03
Refactor Resources

### LINKED ISSUE #606: rustler::resource! produces compiler warning
rust: `1.79.0-nightly (7f2fc33da 2024-04-22)`
rustler: `0.32.1`
elixir: `1.16.2`
erlang: `26.2.4`

this code:
```rust
struct SimplexResource {
    pub gen: Simplex,
}

pub fn load(env: Env, _: Term) -> bool {
    rustler::resource!(SimplexResource, env);
    true
}
```

produces this warning:
```
warning: non-local `impl` definition, they should be avoided as they go against expectation
 --> src/lib.rs:8:5
  |
8 |     rustler::resource!(SimplexResource, env);
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = help: move this `impl` block outside the of the current function `load`
  = note: an `impl` definition is non-local if it is nested inside an item and may impact type checking outside of that item. This can be the case if neither the trait or the self type are at the same nesting level as the `impl`
  = note: one exception to the rule are anon-const (`const _: () = { ... }`) at top-level module and anon-const at the same nesting as the trait or type
  = note: this lint may become deny-by-default in the edition 2024 and higher, see the tracking issue <https://github.com/rust-lang/rust/issues/120363>
  = note: the macro `rustler::resource` may come from an old version of the `rustler` crate, try updating your dependency with `cargo update -p rustler`
  = note: `#[warn(non_local_definitions)]` on by default
  = note: this warning originates in the macro `rustler::resource` (in Nightly builds, run with -Z macro-backtrace for more info)

warning: `noise_nif` (lib) generated 1 warning
```
and with the suggested -Z flag:

```
warning: non-local `impl` definition, they should be avoided as they go against expectation
   --> /home/sk/.local/share/cargo/registry/src/index.crates.io-6f17d22bba15001f/rustler-0.32.1/src/resource.rs:292:13
    |
273 |   macro_rules! resource {
    |   --------------------- in this expansion of `rustler::resource!`
...
292 | /             impl $crate::resource::ResourceTypeProvider for $struct_name {
293 | |                 fn get_type() -> &'static $crate::resource::ResourceType<Self> {
294 | |                     unsafe { &STRUCT_TYPE }.as_ref()
295 | |                         .expect("The resource type hasn't been initialized. Did you remember to call the function where you used the `resource!` macro?")
296 | |                 }
297 | |             }
    | |_____________^
    |
   ::: src/lib.rs:8:5
    |
8   |       rustler::resource!(SimplexResource, env);
    |       ---------------------------------------- in this macro invocation
    |
    = help: move this `impl` block outside the of the current function `load`
    = note: an `impl` definition is non-local if it is nested inside an item and may impact type checking outside of that item. This can be the case if neither the trait or the self type are at the same nesting level as the `impl`
    = note: one exception to the rule are anon-const (`const _: () = { ... }`) at top-level module and anon-const at the same nesting as the trait or type
    = note: this lint may become deny-by-default in the edition 2024 and higher, see the tracking issue <https://github.com/rust-lang/rust/issues/120363>
    = note: the macro `rustler::resource` may come from an old version of the `rustler` crate, try updating your dependency with `cargo update -p rustler`
    = note: `#[warn(non_local_definitions)]` on by default

warning: `noise_nif` (lib) generated 1 warning
```

filmor:
- Move storage of "Type->NIF Resource Type Handle" to a OnceLock map to allow implementing resource types without resorting to dynamic trait implementations
- Implement resource types as a `derive` macro (this one is removed again)
- Add direct access methods to get an immutable reference to the wrapped objects
- Add corresponding converters
- Add explicit resource registration functions
- Add an additional optional "destructor" function to implement use cases like https://github.com/rusterlium/rustler/discussions/626, where a message is supposed to be sent when finalising a resource object
- Implement monitoring support

This allows us to eventually deprecate the existing `resource!` macro and enables a path that fixes #606.


filmor:
Another option, instead of a `derive` macro (or additional to one) would be to allow explicitly implementing a `Resource` trait and registering it using a `rustler::resource!()` outside of `load`, which would use the `inventory::submit!` under the hood. This would allow us to make the `down` callback for #376 (and dyncall and other potential callbacks) a trait function with an empty default implementation.

If we like the `derive` style, another alternative would be, to have a `derive(Resource)` and a `derive(MonitorResource)`, where `MonitorResource` requires an implementation of a `MonitorResource` trait with a `down_callback` function and these would be different code paths with two sets of registrations. I think I'm in favour of that.

filmor:
Note to self: https://github.com/rusterlium/rustler/issues/84 is also a very interesting approach (completely doing away with the necessary pre-registration for resource types!), I'm just not quite sure how to incorporate monitors into it.

scrogson (approved):

PR #623, opened 2024-06-20
Retrieve newest Rustler version from crates.io

filmor:


rustler_mix/lib/rustler.ex
```diff
@@ -170,5 +170,14 @@ defmodule Rustler do
   end
 
   @doc false
-  def rustler_version, do: "0.33.1"
+  def rustler_version do
+    # Retrieve newest version or fall back to hard-coded one
+    Req.get!("https://crates.io/api/v1/crates/rustlaer").body
```
scrogson, line 175: Typo

filmor: Oh yeah, I used that to test the error case :)
Thanks! :)

scrogson (approved):

PR #631, opened 2024-07-07
Implement direct `Result` serialization

filmor:
`serde_rustler` decided to map Rust's `Result<Value, Error>` type to `{ok, Value} | {error, Error}`. While we should keep this functionality, the current implementation is flawed by handling this through actually mapping `ok <-> "Ok"` and `error <-> "Err"`. We can instead detect (de)serialization of a `Result` object by looking at the `name` and `variants` in a variant handler.


filmor:
Superceded by #639.

PR #635, opened 2024-07-11
Add unsafe dyncall support

filmor:
This adds support for dynamic resource calls (https://www.erlang.org/doc/apps/erts/erl_nif#enif_dynamic_resource_call). The feature is inherently unsafe (exchanging pointers to mutable data between different dynamic libraries), though we might be able to eventually provide a stable API on top of it.

Changes to other parts:

- Exposes all of `rustler_sys` under `rustler::sys` s.t. it doesn't have to be explicitly referenced
- Adds helper function to compile tests and nif stubs only for particular NIF versions
- Ensures that `RustlerTest` can be loaded at startup

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ versions.
 - The resource type name can now be overridden with
   `#[register_impl(name = "...")]` (#638)
 - Floats can be decoded from integers (#641, fixes #603)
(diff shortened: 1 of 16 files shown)


filmor:
For reference, the error on Windows is
image

filmor:
The issue described above is fixed with #643

PR #639, opened 2024-07-14
Serde: Restrict special ok/error handling to variant names

benhaney:
This prevents ok/error atoms from getting mangled during transcoding. Also adds a sanity check so only Result variants get special treatment instead of any variant named "Ok" or "Err"

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@ now (#638)
 
 ### Changed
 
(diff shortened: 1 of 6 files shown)


filmor:
I'll add a changelog entry, then this is good to merge, thanks!

filmor (approved):

scrogson (approved):
👍

PR #643, opened 2024-08-05
Fix Windows API functions

filmor:
When adding the newer functions, I messed up the order and missed one function, which breaks rustler-sys for all newer API versions on Windows. Other OS are not affected by this, as they link the symbols directly instead of getting passed a struct.

Making the `rustler_sys` functions `extern "C"` just ensures that they look the same, independent of the operation system.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ versions.
 
 - The optional `register` attribute on `#[register_impl]` works as advertised
   now (#638)
+- API functions for Windows are correctly assigned now for NIF version 2.15 and
+  above (#635)
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


filmor:
@philss @josevalim This finally fixes the issues that we ran into with `enif_term_type`. The name was simply assigned the wrong function because when adding the newer API functions, I wasn't aware of the order-sensitivity :/

josevalim:
@filmor hooray! How did you even find this one out? :D

filmor:
While working on the `dyncall` support in #635, I got really strange errors on Windows when using `enif_init_resource_type` that referenced `enif_term_type`, and then I had a closer look how the API functions are referenced on Windows. :)

PR #644, opened 2024-08-08
Implement Encoder and Decoder for Box<T>

dvic:
I encountered a Clippy warning that my enum got too big and I should box up my inner value, hence the PR ^^.

### DIFF
--- a/rustler/src/types/mod.rs
+++ b/rustler/src/types/mod.rs
@@ -66,6 +66,24 @@ where
     }
 }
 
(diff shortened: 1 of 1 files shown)


filmor:
Maybe expand the `Encoder` for all `Deref<Target=T>`?

dvic:
> Maybe expand the `Encoder` for all `Deref<Target=T>`?

You mean like below? I get a conflicting implementation error like this

image

Issue #648, opened 2024-08-17
Dynamically load Erlang's symbols on startup instead of linking them


filmor:
Doing the "linking" explicitly using `dlopen` and `dlsym` has the advantage that

1. The rustc flags on macOS become unnecessary
2. We can add an additional `main` to list the NIFs without having to fake the symbols (as done in #614)
3. Startup and calling would work the same way on all OSes, allowing us to ditch the build script (and thus speeding up the build)

PR #650, opened 2024-08-19
Dynamic symbols on Unix

filmor:
Implements #648.

We could go even further and drop `rustler-sys` entirely.


filmor:
I looked more into the performance impact, and I think it's compile-time only, at least on Linux. I also compared the assemblies and from what I can see, the generated code for the function calls looks very similar (but still too different to just claim it will execute identically). My current working assumption is that the differences in CI runtimes can be attributed to the CI hosts and potentially a slightly slower startup (as repeated `dlsym` might be slower than linking at load time).

Please review, I will in parallel try to get a pure macro based variant working in another PR.

PR #670, opened 2024-12-17
Elixir 1.18 tweaks and CI

### LINKED ISSUE #669: Typing violation in elixir 1.18.0-rc.0
Compiling a rustler project using the latest elixir 1.18-rc.0 
generates the following compile warning  at the `use Rustler` injection site 

```
    warning: the following clause will never match:

        {^default_load_data_value, {module, function}}

    because it attempts to match on the result of:

        {0, nil}

    which has type:

        {integer(), nil}

    where "default_load_data_value" (context Rustler) was given the type:

        # type: integer()
        # from: lib/rustler_mix_test.ex:1
        default_load_data_value = 0

    typing violation found at:
    │
  1 │ defmodule RustlerMixTest do
    │ ~~~~~~~~~~~~~~~~~~~~~~~~~~~
    │
    └─ lib/rustler_mix_test.ex:1: RustlerMixTest._construct_load_data/0
   ```

The warning originates at this [pattern match](https://github.com/rusterlium/rustler/blob/master/rustler_mix/lib/rustler.ex#L133)
presumably because the compiler infers the types of `load_data` and `load_data_fun` based on the default values set in [struct definition](https://github.com/rusterlium/rustler/blob/master/rustler_mix/lib/rustler/compiler/config.ex#L21)? 🤔 

I'm not sure if it's a false positive in the 1.18 compiler so I opened an issue here first for confirmation 🙇

filmor:
- **Run mix format --migrate to drop usage of 'unless'**
- **Flatten __before_compile__ macro to make the new type check happy** (fixes #669)
- **Add Elixir 1.18 release candidate to CI**

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -85,6 +85,8 @@ jobs:
     strategy:
       matrix:
         pair:
+          - { erlang: "27", elixir: "1.18.0-rc.0" }
+          - { erlang: "26", elixir: "1.18.0-rc.0" }
           - { erlang: "27", elixir: "1.17", latest: true }
           - { erlang: "26", elixir: "1.17" }
           - { erlang: "26", elixir: "1.16" }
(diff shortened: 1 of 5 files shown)


evnu:
LGTM, thanks!

evnu (approved):

PR #676, opened 2025-01-06
Fix Nif* macros requiring an encoder

### LINKED ISSUE #675: Derive expansion error
When using `rustler::NifMap` or `rustler::NifStruct` it is currently not possible to derive only decode. I think this is the same problem as described in #513

The following code fails with `proc-macro derive produced unparsable tokens` and `proc-macro derive produced unparsable tokens`

```rust
#[derive(rustler::NifMap)]
#[rustler(decode)]
struct Test {}
```

If `encode` is used too, the error disappears. The error also does not appear when only `encode` used.

```rust
#[derive(rustler::NifMap)]
#[rustler(decode, encode)]
struct Test {}
```
rustc version: 1.83.0 (90b35a623 2024-11-26)

filmor:
This was probably a left-over from a refactoring. If the encoder was not requested, the macros compiled to code with a dangling `#[allow(...)]` which leads to a parser error.

Fixes #675.

### DIFF
--- a/rustler_tests/native/rustler_test/src/test_codegen.rs
+++ b/rustler_tests/native/rustler_test/src/test_codegen.rs
@@ -225,6 +225,39 @@ pub fn tuplestruct_record_echo(tuplestruct: TupleStructRecord) -> TupleStructRec
     tuplestruct
 }
 
+mod check_if_only_decode_is_enough {
+    // Regression test, failed to compile
+    // TODO: Move this test to the trybuild tests in rustler_codegen
+
+    use rustler::{NifMap, NifRecord, NifStruct, NifTaggedEnum, NifTuple, NifUnitEnum};
+
+    #[derive(NifMap)]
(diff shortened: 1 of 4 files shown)


rustler_tests/native/rustler_test/src/test_codegen.rs
```diff
@@ -225,6 +225,39 @@ pub fn tuplestruct_record_echo(tuplestruct: TupleStructRecord) -> TupleStructRec
     tuplestruct
 }
 
+mod check_if_only_decode_is_enough {
```
evnu, line 228: There is also a test for [encode, decode](https://github.com/rusterlium/rustler/blob/6101dae30bc073f6af6971ac5565f24b6f814e74/rustler_tests/native/rustler_test/src/test_codegen.rs#L31) further above in this file, maybe that test should reference the `mod` here or so..

evnu (approved):

PR #688, opened 2025-03-15
Static NIFs

filmor:
- Still lacks documentation
- The necessary mechanism was released with Erlang 28 (https://www.erlang.org/doc/apps/erts/erl_nif.html#initialization)
- Relevant discussion is https://github.com/rusterlium/rustler/discussions/687.

The general approach is to create a `staticlib` that depends on the relevant NIF libraries (all Rust-based ones have to be linked together) and then passing it into the `configure` script of OTP.


filmor:
@scrogson I'll add docs in #700 and merge as is now.

PR #689, opened 2025-04-10
Replace unsupported definition list syntax in README.md

emig:
**problem**: GitHub Flavored Markdown (GFM) does not support Markdown's definition lists natively.
the current Readme file is rendering the definition list as:

`Type composition : Making a Rust struct encodable and decodable to Erlang or Elixir can be done : with a single attribute.`

**fix**: simulate a definition list using bold terms and line breaks

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,20 +18,17 @@ is favored as of now.
 
 #### Features
 
-Safety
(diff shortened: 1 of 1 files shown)


scrogson:
Thank you.

Issue #685, opened 2025-02-14
Rename libs as we copy them out of target


filmor:
Dropping the lib prefix would make the compiled BEAM modules platform-independent (at least from our side).

filmor:
Also helps with #688

filmor:
Fixed in #691.

PR #691, opened 2025-04-13
Build refactor

filmor:
- **Add staticlib feature to compile without nif_init and add {pkg}_nif_init**
- **Adjust lib name inference**
- **Fix format and remove custom, unreadable env var**
- **Use crate name instead of pkg name**
- **Raise an error if the crate name is not set**
- **Refactor build logic to remove TOML dependency**


scrogson:
@filmor are there other benefits to this? Otherwise, it seems a bit heavy handed if we're just trying to remove a dependency.

filmor:
- It doesn't guess the artifact anymore but instead gets it from Rust's stable interface
- It implements #685, which ensures that the shared library is named the same on all platforms so the `.beam` file becomes platform-independent

PR #692, opened 2025-04-17
improvement(mix): Add `mix rustler.clippy` task.

jimsynz:
This is something I've hacked into a couple of projects for use in our CI pipelines and I thought you might appreciate it being in the main project.


filmor:
Thank you for your contribution! This is not really the direction I have in mind for `rustler_mix`. I'd prefer (and this is also how new projects are created) if rustler projects would use workspaces instead s.t. the same thing can be achieved just by running `cargo clippy` on the root.

PR #694, opened 2025-04-26
Implement Hash for Atom

### LINKED ISSUE #693: HashMap atom keys and accessing atoms
When I pass a Map to Rust it is fine if the keys are Strings or even if the keys and values are Terms.
```rs
#[rustler::nif]
fn call(opts: HashMap<String, String>) -> String {
```

Mostly I try to have the Rust library code with Rust types to let Rustler do the conversion. I have to use Term when the values could be any elixir term, which is probably what it is meant for.

I was wondering why Elixir would have Strings or Symbols as keys but Symbols are not supported as keys in the HashMap in Rust. Values seem to be fine if they are specified as Atoms.
```rs
#[rustler::nif]
fn call(opts: HashMap<Atom, String>) -> String {
```

Produces this:
```
the trait bound `rustler::Atom: Hash` is not satisfied
the trait `Decoder<'_>` is implemented for `HashMap<K, V>`
required for `HashMap<rustler::Atom, rustler::Term<'_>>` to implement `Decoder<'_>`
```

Maybe the keys should never be symbols in Rust, but if this is the case then was looking at the best way to convert. When using Term I can't find an easy way to get an item from the HashMap if wanting to do a string lookup. The best so far is to iterate through the items and convert all symbol keys to strings.

Would symbols as map keys be supported or is there a recommended approach for converting or looking up this type?

scrogson:
This change fixes #693 allowing atoms to be used as keys in a `HashMap`

### DIFF
--- a/rustler/src/types/atom.rs
+++ b/rustler/src/types/atom.rs
@@ -1,6 +1,7 @@
 use crate::wrapper::atom;
 use crate::wrapper::NIF_TERM;
 use crate::{Decoder, Encoder, Env, Error, NifResult, Term};
+use std::hash::{Hash, Hasher};
 
 // Atoms are a special case of a term. They can be stored and used on all envs regardless of where
 // it lives and when it is created.
@@ -96,6 +97,12 @@ impl Encoder for Atom {
 impl<'a> Decoder<'a> for Atom {
     fn decode(term: Term<'a>) -> NifResult<Atom> {
         Atom::from_term(term)
+    }
+}
+
+impl Hash for Atom {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.as_c_arg().hash(state);
     }
 }
 
--- a/rustler_tests/lib/rustler_test.ex
+++ b/rustler_tests/lib/rustler_test.ex
@@ -64,6 +64,7 @@ defmodule RustlerTest do
   def map_from_arrays(_keys, _values), do: err()
   def map_from_pairs(_pairs), do: err()
   def map_generic(_), do: err()
+  def map_atom_keys(_), do: err()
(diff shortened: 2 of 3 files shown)


evnu (approved):
LGTM!

rustler/src/types/atom.rs
```diff
@@ -99,6 +100,12 @@ impl<'a> Decoder<'a> for Atom {
     }
 }
 
+impl Hash for Atom {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.as_c_arg().hash(state);
```
filmor, line 105: Is this guaranteed to work? Wouldn't it be better to use `enif_hash` instead?

evnu: I think this works if we can rely on the NIF_TERM of the same atom being stable across several "instances" of the same atom, right? We are relying on that property for equality checking of atoms as well I believe, so this should hold up here as well. Am I missing something?

But using enif_hash sounds reasonable, as that function knows more about the type.

filmor: The only reason why this would work is that Erlang "inlines" atoms:

https://github.com/erlang/otp/blob/0a3e25826ccbc160cd1dd3866fc3457b9241d8fb/erts/emulator/beam/erl_term.h#L293

As said, I regard this as an implementation detail. There have been multiple discussions over time to loosen the finite/non-gc'd atom table restriction, and if that actually materialized, code relying on the stability of `Eterm` objects would be problematic.

evnu: Sounds reasonable. But will we need to rethink equality as well then? We currently just derive it.

Issue #474, opened 2022-07-07
implement Hash for Atom


dvic:
Can we implement `std::hash::Hash` for `Atom` like this:

```rust
impl std::hash::Hash for Atom {
    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
        state.write_usize(self.term)
    }
}
```

If so, I can open up a PR.

evnu:
Note that the same atom (eg `hello`) could be represented by a different number for different invocations of the VM. Then, the hash for the same atom would also be different. I don't know if this hurts, but it is probably something to think about.

dvic:
Yes and also I now see there is [`Term::hash_internal`](https://docs.rs/rustler/latest/rustler/struct.Term.html#method.hash_internal). And I also see that there is already an implementation for `Term`: 

```rust
impl<'a> Hash for Term<'a> {
    fn hash<H: Hasher>(&self, state: &mut H) {
        // As far as I can see, there is really no way
        // to get a seed from the hasher. This is definitely
        // not optimal, but it's the best we can do for now.
        state.write_u32(self.hash_internal(0));
    }
}
```

So seems like a non-breaking change to add it for the `Atom` type as well, I guess?

evnu:
> So seems like a non-breaking change to add it for the Atom type as well, I guess?

Yes, I think that makes sense. @rusterlium/core opinions? I kinda worry that user's start to write the hash somewhere and assume that it will not change across VM instances.

PR #695, opened 2025-04-27
Implement Hash for atoms

### LINKED ISSUE #693: HashMap atom keys and accessing atoms

filmor:
Alternative to #694. Fixes #693 and #474.

Uses the same has function for `Atom` and `Term`.

The other approach has the significant shortcoming that an atom has a different hash when represented as a `Term` and that it relies on the implementation detail that atom pointers are completely stable.

/edit: Previously also switched to phash2, but given how `std::hash::Hash` is used and documented (not even guaranteed to be stable across Rust versions), it doesn't really help us to pass more "portable" values in.

### DIFF
--- a/rustler/src/types/atom.rs
+++ b/rustler/src/types/atom.rs
@@ -141,7 +141,8 @@ pub(in crate::types) fn decode_bool(term: Term) -> NifResult<bool> {
 impl Hash for Atom {
     fn hash<H: Hasher>(&self, state: &mut H) {
         use crate::sys::{enif_hash, ErlNifHash};
-        let hash = unsafe { enif_hash(ErlNifHash::ERL_NIF_PHASH2, self.as_c_arg(), 0) as u32 };
+        let hash =
+            unsafe { enif_hash(ErlNifHash::ERL_NIF_INTERNAL_HASH, self.as_c_arg(), 0) as u32 };
         state.write_u32(hash);
     }
 }
--- a/rustler/src/types/atom.rs
+++ b/rustler/src/types/atom.rs
@@ -1,6 +1,7 @@
 use crate::wrapper::atom;
 use crate::wrapper::NIF_TERM;
 use crate::{Decoder, Encoder, Env, Error, NifResult, Term};
+use std::fmt;
 use std::hash::{Hash, Hasher};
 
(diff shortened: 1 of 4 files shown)


filmor:
@evnu @scrogson FYI

rustler/src/types/atom.rs
```diff
@@ -137,6 +138,15 @@ pub(in crate::types) fn decode_bool(term: Term) -> NifResult<bool> {
     Err(Error::BadArg)
 }
 
+impl Hash for Atom {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        use crate::sys::{enif_hash, ErlNifHash};
+        let hash =
+            unsafe { enif_hash(ErlNifHash::ERL_NIF_INTERNAL_HASH, self.as_c_arg(), 0) as u32 };
```
evnu, line 145: Question: This somewhat duplicates the code in Term, should we typecast somehow to reuse the same code? Or do we need to keep the code separate?

scrogson (approved):

PR #704, opened 2025-09-09
Add from_iter to NewBinary

aDifferentJT:
#702 added conformance to `FromIterator` to `OwnedBinary`, `NewBinary` can't do the same since it requires an `Env` and lacks realloc, however a function that does similarly, though with the added requirement of `ExactSizeIterator` would be useful.


filmor:
Thanks! Would you mind adding tests for this and #702?

PR #724, opened 2026-04-12
Make `OwnedEnv` `Sync`-compatible

hauleth:
It should be perfectly fine to do, in the same way #469 makes
`OwnedBinary` `Sync`-compatible.

I require it to make `OwnedEnv` part of the structure that is storing
`Term`s within its body, just like it happens with ETS.

### DIFF
--- a/rustler/src/env.rs
+++ b/rustler/src/env.rs
@@ -206,6 +206,7 @@ pub struct OwnedEnv {
 }
 
 unsafe impl Send for OwnedEnv {}
+unsafe impl Sync for OwnedEnv {}
 
 impl OwnedEnv {
     /// Allocates a new process-independent environment.


filmor:
Quoting from the docs (https://www.erlang.org/doc/apps/erts/erl_nif.html):

> A NIF is thread-safe without any explicit synchronization as long as it acts as a pure function and only reads the supplied arguments. When you write to a shared state either through static variables or [enif_priv_data](https://www.erlang.org/doc/apps/erts/erl_nif.html#enif_priv_data), you need to supply your own explicit synchronization. **This includes terms in process independent environments that are shared between threads.** Resource objects also require synchronization if you treat them as mutable.

I don't think process-independent environments as such are thread-safe "enough" for `Sync`.

hauleth:
@filmor quoting docs:

> Created by calling `enif_alloc_env`. This environment can be **used to store terms between NIF calls** and to send terms with `enif_send`. A process independent environment with all its terms is valid until you explicitly invalidate it with `enif_free_env` or `enif_send`.

Emphasis mine. Rustler requires that the value returned from the NIF call must be `Sync`, which mean that `OwnedEnv` must be `Sync` to be used in such situation.

filmor:
Sorry for the high latency :)

I am not quite following what you are trying to do that would require `OwnedEnv` to be `Sync`. Erlang terms _are_ `Sync`, so all terms living in an `OwnedEnv` can be passed back from the NIF (though that will in any case first copy the term back into the process environment).

My point was that `ErlNifEnv` _as such_ might not be thread-safe enough. It does say that this is only relevant for writing, so it could be that the same logic applies as in #469 (Rust's rules around `&mut T`).

hauleth:
Sorry for higher latency - the idea is to store terms inside of the NIF. Imagine something like ETS implemented in Rust - we need to keep the `OwnedEnv` to keep the value of the items in the DB alive.

filmor:
But why do you need it to be `Sync` for this? Put it into a `Arc<Mutex<>>` and grab a lock whenever you need it (e.g. to copy a term in). The `Sync` was explicitly removed in https://github.com/rusterlium/erlang_nif-sys/issues/25, and I agree with @goertzenator's and @jorendorff's assessment, that the docs read like `ErlNifEnv` not being thread-safe (in the sense of allowing fully concurrent usage).

Tracing through the code of (e.g.) `enif_make_copy`, we go via `make_copy` to `alloc_heap` to (eventually) [`alloc_heap_heavy`](https://github.com/erlang/otp/blob/12c91e62d4c4dfd94ff413d22a52272a0a5b6760/erts/emulator/beam/erl_nif.c#L217-L237), and this one is quite obviously not thread-safe, modifying `env`-attached structures without locking.

Thus I'm strongly leaning towards closing this PR.

PR #726, opened 2026-05-14
nif_filler: fix Android NIF init panic from dlopen(NULL) lookup

GenericJam:
## Bug

When BEAM is statically linked into an Android app and a rustler NIF is loaded, NIF init panics with:

```
thread '<unnamed>' panicked: 'DlSym { source: "undefined symbol: enif_priv_data" }'
```

The panic comes from `sys::nif_filler` calling `Library::open(None, ...)` (i.e. `dlopen(NULL)`) to find `enif_*` symbols, then `dlsym`-ing each one off the resulting handle.

## Root cause

Bionic's `dlopen(NULL)` returns a handle to the main executable's link namespace — on a JVM-hosted Android app, that's `app_process`. That namespace does not resolve symbols from sibling `.so` libraries loaded via `System.loadLibrary`, even when the consumer has subsequently re-`dlopen`'d the app's own `.so` with `RTLD_NOW | RTLD_GLOBAL` from JNI/host code.

I tried `RTLD_DEFAULT` first under the assumption that it would traverse the global symbol scope; it has the same blind spot on Bionic. Empirically, the only way to get `dlsym` to resolve the statically-linked-in `enif_*` exports on Android is to open an **explicit handle to the `.so` that contains rustler itself**.

## Fix

Identify the `.so` containing rustler via `dladdr` of one of our own functions, then `dlopen(self_path, RTLD_NOW | RTLD_NOLOAD)` for an explicit handle (`NOLOAD` so we don't reload the library, just return a handle to the existing load). `dlsym` against that handle resolves the `enif_*` symbols correctly.

The change is gated behind `#[cfg(target_os = "android")]`. Linux, macOS, iOS, and Windows code paths are untouched.

Additional small fix: rustler's caller passes symbol names with a trailing NUL byte already (e.g. `b\"enif_priv_data\\0\"`), so we pass `bytes.as_ptr() as *const c_char` to `dlsym` directly instead of going through `CString::new(name).unwrap()` (which rejects the trailing NUL as "interior NUL"). A `debug_assert!` confirms the trailing NUL is present.

## Where this is in production

We use this in [Mob](https://github.com/GenericJam/mob), a framework for shipping BEAM apps on iOS and Android. Mob statically links BEAM + rustler into the main Android `.so`. Verified end-to-end on an Android emulator and a physical Pixel — three concurrent NIFs (C, Rust, Zig) statically linked into a single app, all initialise and respond to Erlang RPC.

Tracker on our side: [GenericJam/mob#7](https://github.com/GenericJam/mob/issues/7).

## Notes

- The new Android module panics with a descriptive message on `dladdr` / `dlopen` failure (path + dlerror). Those cases shouldn't occur in practice (we're identifying our own `.so`), but a loud panic seems better than silent symbol-table corruption.
- Happy to iterate on the shape — e.g. if you'd prefer a single-module implementation with `cfg(target_os = "android")` branches rather than a separate Android `internal` module, or want a different panic message style.

Edit by human: This is generated by AI (Claude Opus 4.6).
I understand if you don't accept AI PRs but this is what Mob needs for Rust / Rustler to work.
Feel free to pick from this what is genuinely useful as I would prefer to refer to the main repo than my fork.

How it's being used in Mob: https://hexdocs.pm/mob_dev/nifs.html#rust-via-rustler

### DIFF
--- a/rustler/src/sys/nif_filler.rs
+++ b/rustler/src/sys/nif_filler.rs
@@ -2,7 +2,7 @@ pub(crate) trait DynNifFiller {
     fn write<T: Copy>(&self, field: &mut Option<T>, name: &str);
 }
 
-#[cfg(not(target_os = "windows"))]
+#[cfg(all(not(target_os = "windows"), not(target_os = "android")))]
 mod internal {
     use std::ffi::OsStr;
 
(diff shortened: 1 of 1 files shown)


filmor:
1. I am open to the change and the approach
2. I would suggest that, if you haven't yet, give https://github.com/rusterlium/rustler/discussions/687 a read, especially regarding linking multiple Rust nif libs with LTO

GenericJam:
Read through #687 and the merged PR #688
 
 - `<crate>_nif_init` always emitted, matching the OTP 28 static-NIF table's lookup symbol
 - `bare nif_init` only when `CARGO_PRIMARY_PACKAGE` or `RUSTLER_PRIMARY_NIF_INIT` is set, which removes the duplicate-symbol-from-two-Rust-crates collision adiibanez hit

The LTO / Rust-runtime duplicate-symbol issue (`EMPTY_PANIC`, `rust_eh_personality`) is a downstream-project concern; we set `-C lto=off -C embed-bitcode=no` on the bundling side and that's enough.
 
This PR is orthogonal — it's the run-time complement to #688. After `<crate>_nif_init` runs, `internal_write_symbols()` populates the dyn-callback table via `nif_filler`, which currently does `dlopen(NULL)` on non-Windows. Bionic's `dlopen(NULL)` returns the `app_process` link namespace, which doesn't resolve `enif_*` symbols statically linked into a sibling .so even when that .so was loaded by System.loadLibrary with `RTLD_NOW` | `RTLD_GLOBAL`. The fix uses `dladdr` to identify the .so containing rustler and `dlopen(self_path, RTLD_NOW | RTLD_NOLOAD)` to get an explicit handle that dlsym does resolve `enif_*` against. Android-only `#[cfg]`, no behaviour change anywhere else.
 
Pushed a cargo fmt fix just now (c4f706e); the earlier red was only the Format job — every other failure on the run was a cancelled downstream of that.

One thing #688 doesn't quite cover for our use case in Mob: the recipe is "build N per-NIF staticlibs, link them all into one app  .so", and cargo sets `CARGO_PRIMARY_PACKAGE=1` for every leaf staticlib build. So each per-NIF .a carries both `<crate>_nif_init` and a `bare nif_init`, and linking two of them back collides again. The bundling-staticlib pattern from the discussion sidesteps it, but at the cost of having a separate bundling crate per project that re-exports everything. Would you be open to a follow-up that adds a `RUSTLER_NO_PRIMARY_NIF_INIT=1` (or similarly-named) escape hatch that forces-off the bare emission even when `CARGO_PRIMARY_PACKAGE` is set? Happy to do that as a separate PR; would keep this one narrowly about the Android dlsym fix.

Human edit: most of this is coming out of Claude. I roughly understand but bring little expertise to the table wrt Rust. I'm doing this to make it available to Mob users.

filmor:
- **When I ask you to please read through a discussion, that means I want _you_ to read it, not a machine**
- In the discussion you can see that there are serious issues with simply linking together multiple statically compiled Rust libraries, the LTO one is not fixed
- The correct way to go about this is to introduce a custom staticlib crate that bundles all Rustler NIFs (can also include C or Zig NIFs, if you want that)
- Your agent's whole understanding of how `nif_init` works is spotty. The `nif_init` symbol isn't some decision that the library takes, it is what the BEAM uses when the respective module is loaded as a shared library.
- I have tested this quite thoroughly, the `nif_init` symbol only appears in a primary crate while the `<libname>_nif_init` symbol is always generated

In summary:
- The idea implemented in this PR is probably fine
- If you want to interact with me, do so by painstakingly writing yourself, please
- Please consider the combined staticlib approach, there is a reason why we landed on that

GenericJam:
> * If you want to interact with me, do so by painstakingly writing yourself, please

I copied the technical part above but I did actually hand edit it so I wasn't `copy/paste`ing Claude's text verbatim. Claude didn't bother giving you backticks and the linebreaks were 80 chars wide.

btw Mob is a solo effort by me so if I didn't have Claude helping it just _wouldn't_ happen at all.

I'm trying to make Rust and Elixir more accessible and useful on mobile.

I do appreciate your work with this library. I've used it in the past to great effect even though I've never made it as far as github so thank you for your contribution to the Rust/Elixir ecosystem.

Thank you for the tip about the combined static lib. That would have shown up as a bug in the future so that steered us around some potholes.

filmor:
This is a relevant comment on the issue: https://github.com/android/ndk/issues/201

For backwards-compatibility with its old, broken implementation of `RTLD_GLOBAL`, Android's linker will only consider symbols part of `RTLD_GLOBAL` that are included in the main executable.

The given implementation has the major disadvantage that it will _only_ work if the `enif_*` symbols are actually contained in the same `.so` as the Rustler library. While this is a perfectly valid approach, it is not a generic solution, one might want to run a non-statically linked BEAM in other setups.

Is there a way for us to know the path from within Mob and set it as an environment variable? Then we could simply (and generically) extend the `DlsymNifFiller` to use that one instead.

GenericJam:
The mechanism is conclusively proven on a physical arm64 Android device (moto 5 2021):

`env var` contract works end-to-end with a real Rust NIF executing (as per multi NIF demo)

`dladdr` fallback preserved unchanged (you can verify by unsetting the env var and re-running; it falls back to the `dladdr-self` path, same behaviour as the existing PR)

No iOS / non-Android / Windows changes

No further changes needed on rustler end beyond this PR. It's a config change on the Mob side. Thanks for the heads up!

filmor:
Can you please test the change from #733? It _should_ be equivalent to the implementation you added here (slightly different choice of env variable name, `RUSTLER_BEAM_LIBRARY_PATH`), but a lot smaller and without the `dladdr` codepath, that I don't want to maintain.

GenericJam:
Tested #733 against Mob on Android arm64. Works the same as #726 for Mob. Happy to close #726. Thanks for your help!

filmor:
Great, thanks for pointing out the issue. The other MR is merged now and I'll probably release 0.38 on the weekend.

Issue #727, opened 2026-05-14
Support renamed derive fields and a term_map! helper for atom-keyed maps


dannote:
Rustler users often expose Elixir-facing APIs made of plain maps with atom keys. Those keys should follow Elixir naming conventions, not Rust identifier constraints.

A common issue is that natural Elixir map keys include names that are Rust keywords or otherwise awkward Rust identifiers:

```elixir
%{
  type: :import,
  start: 0,
  end: 12,
  async: true
}
```

In Rust, these fields usually need different names:

```rust
struct ImportInfo {
    type_: rustler::Atom,
    start: u32,
    end_: u32,
    async_: bool,
}
```

Without a way to rename the derived atom names, users have to either change the Elixir API shape or write manual `Encoder` / `Decoder` implementations.

Two small additions would cover this pattern well:

1. `#[rustler(rename = "...")]` support for derived field / variant atom names.
2. A `term_map!` macro for constructing Erlang maps without parallel key/value arrays.

## `#[rustler(rename = "...")]`

With rename support, users could keep idiomatic Elixir keys while still using valid Rust identifiers:

```rust
#[derive(NifMap)]
struct ImportInfo {
    #[rustler(rename = "type")]
    type_: rustler::Atom,

    start: u32,

    #[rustler(rename = "end")]
    end_: u32,

    #[rustler(rename = "async")]
    async_: bool,
}
```

This would encode/decode:

```elixir
%{
  type: :import,
  start: 0,
  end: 12,
  async: true
}
```

This mirrors a familiar Rust derive convention from Serde:

```rust
#[serde(rename = "type")]
type_: String
```

Potential scope:

- `NifMap` field names
- `NifTaggedEnum` named variant fields
- `NifTaggedEnum` variant tags
- `NifUnitEnum`, `NifStruct`, and `NifRecord` for consistency

Starting with only `NifMap` fields would already solve the most common plain-map case.

## `term_map!`

Rustler already exposes the necessary low-level API for map construction, but common map construction currently requires parallel key/value arrays:

```rust
rustler::Term::map_from_term_arrays(
    env,
    &[
        atoms::code().encode(env),
        atoms::css().encode(env),
        atoms::errors().encode(env),
        atoms::warnings().encode(env),
    ],
    &[
        code.encode(env),
        css.encode(env),
        errors.encode(env),
        warnings.encode(env),
    ],
)
.unwrap()
```

For larger maps, the key and value for a field are separated, which makes additions and reordering more error-prone.

[...]

PR #728, opened 2026-05-14
Add rename support for derived atom names

dannote:
Adds `#[rustler(rename = "...")]` support for derive-generated atom names.

This lets Rust fields and enum variants keep valid Rust identifiers while encoding/decoding idiomatic Elixir keys like `:type`, `:end`, or `:async`.

Covered cases:

- `NifMap` field names
- `NifTaggedEnum` named fields
- `NifTaggedEnum` variant tags

The rename parser rejects malformed rename attributes, and the tests cover both runtime behavior and invalid syntax.

Refs #727.

Tests run:

- `cargo test -q`
- `cd rustler_tests && mix test test/codegen_test.exs`


rustler_codegen/src/context.rs
```diff
@@ -110,29 +110,48 @@ impl<'a> Context<'a> {
...
+
+    pub fn ident_to_atom_name(ident: &Ident) -> String {
+        let ident_str = ident.to_string();
+        Self::remove_raw(&ident_str).to_snake_case()
+    }
+
+    fn atom_fun(atom_name: &str) -> Ident {
+        let suffix = atom_name
+            .chars()
+            .map(|c| if c.is_ascii_alphanumeric() { c } else { '_' })
```
filmor: This is pretty aggressive and can fail with a confusing error in cases like `name_ä` vs `name_ö`. Wouldn't it make more sense to just fail with a proper error here? Apart from that, atoms _can_ be at least latin1, on newer versions utf8.

PR #731, opened 2026-05-17
Add OTP-29 enif functions

filmor:
The two atom cache functions are not yet documented, as far as I can see. We add them here anyways to not later trip up the Windows build as the API stubs struct has to be complete.


evnu (approved):
LGTM. Curious why the atom cache functions aren't documented.

Issue #645, opened 2024-08-10
Use UTF8 functions


filmor:
NIF version 2.17 (OTP 26) got new functions and extended existing functions around strings and atoms to support UTF8 encoded strings directly. As Rust `str` uses UTF8, this should allow us to simplify the code and potentially also speed things up, as we can skip intermediate allocations.

PR #732, opened 2026-05-17
Implement UTF8 atom support on NIF 2.17

filmor:
- Add `(try_)from_utf8_bytes` on all NIF versions, use builtin fast path
  on NIF 2.17 upwards
- Use direct UTF8 conversions for strings

Implements #645

### DIFF
--- a/rustler/src/types/atom.rs
+++ b/rustler/src/types/atom.rs
@@ -23,12 +23,12 @@ impl Atom {
     }
 
     unsafe fn from_nif_term(term: NIF_TERM) -> Self {
-        Atom { term }
+        Self { term }
     }
 
     pub fn from_term(term: Term) -> NifResult<Self> {
         if term.is_atom() {
-            Ok(unsafe { Atom::from_nif_term(term.as_c_arg()) })
+            Ok(unsafe { Self::from_nif_term(term.as_c_arg()) })
         } else {
             Err(Error::BadArg)
         }
@@ -41,7 +41,7 @@ impl Atom {
     ///
     /// This function uses Latin-1 encoding for compatibility.
     pub fn from_bytes(env: Env, bytes: &[u8]) -> NifResult<Self> {
-        Atom::from_encoded_bytes(env, bytes, ErlNifCharEncoding::ERL_NIF_LATIN1)
+        Self::from_encoded_bytes(env, bytes, ErlNifCharEncoding::ERL_NIF_LATIN1)
     }
 
     /// Return the atom whose text representation is UTF-8 `bytes`.
@@ -57,7 +57,7 @@ impl Atom {
 
         #[cfg(feature = "nif_version_2_17")]
         {
-            Atom::from_encoded_bytes(env, bytes, ERL_NIF_UTF8)
+            Self::from_encoded_bytes(env, bytes, ERL_NIF_UTF8)
         }
 
         #[cfg(not(feature = "nif_version_2_17"))]
@@ -70,13 +70,13 @@ impl Atom {
                 }
                 latin1.push(c as u8);
             }
-            Atom::from_encoded_bytes(env, &latin1, ERL_NIF_LATIN1)
+            Self::from_encoded_bytes(env, &latin1, ERL_NIF_LATIN1)
         }
     }
 
-    fn from_encoded_bytes(env: Env, bytes: &[u8], encoding: ErlNifCharEncoding) -> NifResult<Atom> {
+    fn from_encoded_bytes(env: Env, bytes: &[u8], encoding: ErlNifCharEncoding) -> NifResult<Self> {
         unsafe {
-            atom::make_atom(env.as_c_arg(), bytes, encoding).map(|term| Atom::from_nif_term(term))
+            atom::make_atom(env.as_c_arg(), bytes, encoding).map(|term| Self::from_nif_term(term))
         }
     }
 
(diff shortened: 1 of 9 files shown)


rustler_tests/test/atom_test.exs
```diff
@@ -7,6 +7,24 @@ defmodule RustlerTest.AtomTest do
...
 
+  test "utf8 atom roundtrip" do
+    utf8 = "ÀrgerÖ"
+
+    utf8_atom = RustlerTest.binary_to_atom_utf8(utf8)
+    assert utf8_atom == String.to_atom(utf8)
+    assert RustlerTest.atom_to_string(utf8_atom) == utf8
+  end
+
+  test "utf8 atom roundtrip on nif 2.17+" do
```
evnu, line 18: **question** why an extra test for 2.17, shouldn't the fallback code also be able to cover this case?

rustler_tests/lib/rustler_test.ex
```diff
@@ -95,7 +95,9 @@ defmodule RustlerTest do
   def atom_to_string(_), do: err()
   def atom_equals_ok(_), do: err()
   def binary_to_atom(_), do: err()
+  def binary_to_atom_utf8(_), do: err()
   def binary_to_existing_atom(_), do: err()
+  def binary_to_existing_atom_utf8(_), do: err()
```
evnu, line 102: **nitpick** this function is not called yet.

rustler/src/wrapper/atom.rs
```diff
@@ -1,22 +1,49 @@
...
     enif_make_atom_len(env, name.as_ptr() as *const c_char, name.len())
 }
 
-pub unsafe fn make_existing_atom(env: NIF_ENV, name: &[u8]) -> Option<NIF_TERM> {
+#[cfg(feature = "nif_version_2_17")]
+pub unsafe fn make_atom(env: NIF_ENV, name: &[u8], encoding: ErlNifCharEncoding) -> NIF_TERM {
+    let mut atom_out: NIF_TERM = 0;
+
+    // Create a new atom with the requested encoding.
+    // Returns 0 if creation fails (e.g. invalid text/encoding).
```
evnu, line 33: **suggestion** can we test the error case?

evnu (approved):
LGTM. Left a few questions, nothing blocking from my POV.

rustler_tests/test/atom_test.exs
filmor: Before 2.17 this would fail as the respective codepoints can't be encoded in latin1.

rustler_tests/lib/rustler_test.ex
filmor: True, I'll add a test.

rustler_tests/test/atom_test.exs
evnu: Yeah, that makes sense.. Time for coffee. :)

filmor:
@evnu Can you re-review? I have added the tests, but while doing so, I recognized that we can't really distinguish between "atom string is broken" and "atom does not exist". Erlang doesn't distinguish these cases either. I have thus modified the interface to return a bare `NifResult<Atom>` instead of `NifResult<Option<Atom>>`. This _is_ a breaking change (added it to the UPGRADE.md) and it is not strictly necessary, but it's definitely the better choice in the long run IMO.

rustler/src/types/atom.rs
```diff
@@ -33,54 +34,125 @@ impl Atom {
...
+    /// if atom with given text representation exists.
+    ///
+    /// # Errors
+    /// `Error::BadArg` if the bytes are incorrectly encoded, the array is too long (255 bytes before
+    /// NIF 2.17, 255 characters later), or the atom already exists.
+    fn try_from_encoded_bytes(
+        env: Env,
+        bytes: &[u8],
+        encoding: ErlNifCharEncoding,
+    ) -> NifResult<Atom> {
```
evnu: **suggestion** Maybe also rename `Atom` to `Self` (similar cases further below as well). Just a nitpick, but we have that above as well.

```suggestion
    ) -> NifResult<Self> {
```

evnu (commented):
> I have added the tests, but while doing so, I recognized that we can't really distinguish between "atom string is broken" and "atom does not exist". 

In that case, does keeping `try_` even make sense? The code is identical now.

filmor:
> > I have added the tests, but while doing so, I recognized that we can't really distinguish between "atom string is broken" and "atom does not exist".
> 
> In that case, does keeping `try_` even make sense? The code is identical now.

It's not. One is equivalent to `binary_to_atom`, the other to `binary_to_existing_atom`. We _want_ people to use `binary_to_existing_atom` (so `try_from_str` etc.), and this removes some clunkiness from the API.

evnu:
> One is equivalent to `binary_to_atom`, the other to `binary_to_existing_atom`.

Sorry I missed that before. In that case, I think your suggestion makes sense! I wonder if we should encode in the `try_` function name that the function is trying for an existing atom. I don't want to bike shed here though.

filmor:
Good point. That would also allow us to keep the interface of the "old" function in-tact for a while and mark it as deprecated.

filmor:
- Added a new set of names `existing_from...`
- Reverted `try_from_bytes` and `try_from_str` to the old behaviour
- Marked both as deprecated

PR #737, opened 2026-05-22
Document the lib configuration option

### LINKED ISSUE #391: Document `lib` configuration option in `Rustler`?
The `lib` option is used by the `binary_example` test NIF, but it is not yet documented. Maybe we should document it?

wyf027:
## Summary
- document the `:lib` Rustler configuration option
- mention its default and when to set it to `false` for binary-only crates

Closes #391

## Verification
- `git diff --check`
- `cargo check -p rustler --manifest-path /Users/wuyangfan/rustler-contrib/Cargo.toml`

## Risk
- Documentation-only change; no runtime behavior changes.


filmor:
Seeing that I'd rather remove this feature in the future, I don't think we should document it.

PR #740, opened 2026-05-22
Implement registering and calling stop callback of resources

### LINKED ISSUE #735: Add stop callback support for resources

lilioid:
I have implemented a new `stop()` callback compatible with [ErlNifResourceStop](https://www.erlang.org/doc/apps/erts/erl_nif.html#ErlNifResourceStop) as an addition to the `Resource` trait.

This has been manually verified to work by implementing a small program on top of it.
The codegen for the `#[resource_impl]` macro also just works.

resolves #735

### DIFF
--- a/rustler/src/resource/registration.rs
+++ b/rustler/src/resource/registration.rs
@@ -3,8 +3,9 @@ use super::util::align_alloced_mem_for_struct;
 use super::ResourceInitError;
 use crate::env::EnvKind;
 use crate::sys::{
-    c_char, c_void, ErlNifEnv, ErlNifMonitor, ErlNifPid, ErlNifResourceDown, ErlNifResourceDtor,
-    ErlNifResourceFlags, ErlNifResourceType, ErlNifResourceTypeInit,
+    c_char, c_int, c_void, ErlNifEnv, ErlNifEvent, ErlNifMonitor, ErlNifPid, ErlNifResourceDown,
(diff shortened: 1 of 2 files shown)


filmor:
Thank you for this, I will incorporate these changes (sans the odd `nif_version` flag adjustment that I'm pretty certain is wrong...) into #738 which should implement the actual `enif_select` wrapper.

lilioid:
> sans the odd nif_version flag adjustment that I'm pretty certain is wrong...

Ahh yeah that must have slipped through from when I was confused about OTP vs. NIF versions (hence the #739 ).

> Thank you for this, I will incorporate these changes into #738 which should implement the actual `enif_select` wrapper.

A safe wrapper is definitely the preferred way to go and whichever way you want to incorporate this works for me. This was just what I needed to get my things running :sweat_smile:

PR #743, opened 2026-05-25
Drop build.rs in favour of macro_rules

filmor:
Experiment, pointed Copilot at the `build.rs` to see whether we can get rid of it. I tried this before by hand and got quite far, but these recursive macro invocations are really annoying to write :)


filmor:
Closed, because we might still require `build.rs` in the future anyhow to check for `panic=abort`

PR #746, opened 2026-06-12
Support NifMap decoding when elixir map has missing fields

### LINKED ISSUE #745: Support NifMap decoding when elixir map has missing fields
Currently, if we derive `NifMap` for a struct like the following, attempts to decode `%{}` will fail, because the derived code expects all fields to exist in the map:

```rust
#[derive(NifMap)]
#[rustler(decode)]
struct MaybeMissingField<'a> {
    field: Option<&'a str>,
}
```

The "workaround" code that I found for this involves using a `HashMap<Atom, Term>` instead of a struct with `NifMap` derived, which is definitely not ideal.

`NifMap` could support decoding from maps that have missing fields, as long as the type of each missing field is an `Option`. Perhaps this behaviour could be gated behind an additional attribute, either on the struct or the field, to allow missing fields be represented as `None`.

I can work on a PR for this, but I decided to open an issue first to understand if this is something desirable, and to work out the details of the implementation (the potential attributes mentioned above).

DanielSidhion:
Closes #745.

This PR currently only implements the behaviour for `NifMap`, but it could in theory support `NifStruct` as well, although the utility of that is way lower, since elixir structs will always have the fields - but in Rust they could map to a struct with more fields that will always become `None`.

It also only implements the decoding behaviour, but encoding could be done similarly: if the option is `None`, don't add that key to the elixir map. I could implement that for completeness.

The PR has 3 commits, it might be easier to review commit by commit to understand the changes faster.
The implementation is not ideal (e.g. the way I check whether a type is `Option` has false negatives and false positives, as explained in the comments), but given the alternative (explained later) and how the rest of the code behaves, I decided this was a good approach to take.

Instead of checking if the type of a field is `Option`, we could've done the entire implementation at runtime: when the elixir map doesn't have the field we're looking for, delegate the return value to a trait. This trait would return `Ok(None)` for `Option<T>` types, and and `Err(_)` value for all other types. However, this would require some sort of extra trait to gate the implementation specifically for `Option<T>`, because we can't just have an implementation for `T` and `Option<T>`. My idea here would be to add an extra "marker" trait for each type that implements `rustler::Decoder` except for `Option<T>`. This would be a lot of overhead compared to just checking the type at procgen time, so I opted for that approach instead.

### DIFF
--- a/rustler_codegen/src/attrs.rs
+++ b/rustler_codegen/src/attrs.rs
@@ -36,6 +36,7 @@ pub(crate) trait TryFromRustlerNestedAttr: Sized {
 pub(crate) enum RustlerAttr {
     Encode,
     Decode,
+    OptionalDecode,
     Module(String),
     Tag(String),
 }
@@ -71,7 +72,7 @@ impl RustlerAttr {
 
 impl TryFromRustlerNestedAttr for RustlerAttr {
     fn parse_failure_message() -> impl Display {
-        "Expected encode, decode and/or optional in rustler attribute"
+        "Expected encode, decode and/or optional_decode in rustler attribute"
     }
 
     fn collect_attrs_for_ident(ident: &Ident, meta: &Meta) -> Option<Vec<Self>> {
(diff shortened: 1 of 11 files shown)


rustler_codegen/src/attrs.rs
```diff
@@ -0,0 +1,93 @@
+use std::fmt::Display;
+
+use syn::{Ident, Lit, Meta};
+
+pub(crate) trait TryFromRustlerNestedAttr: Sized {
```
DanielSidhion, line 5: The name for this trait should change to be a bit more global than just "rustler nested attributes", but I couldn't think of anything interesting. Appreciate any suggestions.

PR #749, opened 2026-06-18
Go back to pre-generating all API snippets

filmor:
- Move generator code out of build.rs into a separate, unpublished crate
- Adjust the generator to use `lang_c` for parsing declaration files
  instead of hand-copying the changes
- Use the original `erl_nif_api_funcs.h` files as as the basis for this
  process


filmor:
The only differences in the APIs are:

- Missing APIs added (apart from the thread APIs that have very limited use in Rust), in particular the IO parts
- The atom APIs now use `c_char` instead of `c_uchar`, in line with the headers

PR #756, opened 2026-08-21
Add hashing for LocalPid

### LINKED ISSUE #755: Implement `Hash` for `LocalPid`
Sometimes it may be useful to use `LocalPid` as key in some map or set, currently we can use only `BTree*` container family, because `LocalPid` is sortable, but not hashable.

evnu:
Use `ERL_NIF_INTERNAL_HASH` to generate a hash in a similar fashion as we do for `Atom`.

Fix #755

### DIFF
--- a/rustler/src/types/local_pid.rs
+++ b/rustler/src/types/local_pid.rs
@@ -2,6 +2,7 @@ use crate::sys::{enif_compare_pids, enif_is_process_alive, enif_self};
 use crate::wrapper::{pid, ErlNifPid};
 use crate::{Decoder, Encoder, Env, Error, NifResult, Term};
 use std::cmp::Ordering;
+use std::hash::{Hash, Hasher};
 use std::mem::MaybeUninit;
 
 #[derive(Copy, Clone)]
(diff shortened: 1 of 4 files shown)


rustler/src/types/local_pid.rs
```diff
@@ -88,3 +89,11 @@ impl Env<'_> {
         res != 0
     }
 }
+
+impl Hash for LocalPid {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        use crate::sys::{enif_hash, ErlNifHash};
+        let hash = unsafe { enif_hash(ErlNifHash::ERL_NIF_INTERNAL_HASH, self.as_c_arg().pid, 0) };
```
evnu, line 96: Accessing the internal `pid` field of `ErlNifPid` is a bit icky. Is there a better way? We need an `ERL_NIF_TERM` for `enif_hash`.

filmor: I doubt there is without an `Env`. You _could_ create an `OwnedEnv` and "move" the pid into that with `encode`, but maybe we do that when `ErlNifPid` _actually_ changes :)

PR #757, opened 2026-08-21
Implement `From<&[u8]>` for binary

hauleth:
This allows for simpler work with binaries that aren't UTF-8 encoded
strings. Currently there is either need to construct such data in place
(which limits usage of some Rust libraries) or you need to write
additional code to translate one form into another. This implementation
is simple and performant enough for this usecase.

### DIFF
--- a/rustler/src/types/binary.rs
+++ b/rustler/src/types/binary.rs
@@ -262,6 +262,17 @@ impl FromIterator<u8> for OwnedBinary {
     }
 }
 
+impl<T: AsRef<[u8]>> From<T> for OwnedBinary {
+    fn from(data: T) -> Self {
+        let buf = data.as_ref();
+        let mut bin = Self::new(buf.len()).expect("Allocation failed");
+
+        bin.as_mut_slice().copy_from_slice(buf);
+
+        bin
+    }
+}
+
 /// An immutable smart-pointer to an Erlang binary.
 ///
 /// See [module-level doc](index.html) for more information.
--- a/rustler/src/types/binary.rs
+++ b/rustler/src/types/binary.rs
@@ -135,6 +135,14 @@ impl OwnedBinary {
         })
     }
 
+    /// Copies 'data''s data into a new `OwnedBinary`.
+    pub fn from_slice(data: impl AsRef<[u8]>) -> Self {
+        let data = data.as_ref();
+        let mut bin = OwnedBinary::new(data.len()).expect("allocation failed");
+        bin.as_mut_slice().copy_from_slice(data);
+        bin
+    }
+
     /// Attempts to reallocate `self` with the new size.
     ///
     /// Memory outside the range of the original binary will not be initialized. If
@@ -264,12 +272,7 @@ impl FromIterator<u8> for OwnedBinary {
 
 impl<T: AsRef<[u8]>> From<T> for OwnedBinary {
     fn from(data: T) -> Self {
-        let buf = data.as_ref();
-        let mut bin = Self::new(buf.len()).expect("Allocation failed");
-
-        bin.as_mut_slice().copy_from_slice(buf);
-
-        bin
+        Self::from_slice(data)
     }
 }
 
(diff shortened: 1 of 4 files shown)


rustler/src/types/binary.rs
```diff
@@ -262,6 +262,16 @@ impl FromIterator<u8> for OwnedBinary {
     }
 }
 
+impl From<&[u8]> for OwnedBinary {
+    fn from(buf: &[u8]) -> Self {
+        let mut bin = Self::new(buf.len()).expect("Allocation failed");
```
filmor: This should not panic, please implement `TryFrom` instead.

rustler/src/types/binary.rs
```diff
@@ -262,6 +262,16 @@ impl FromIterator<u8> for OwnedBinary {
...
+
+        bin.as_mut_slice().copy_from_slice(buf);
```
filmor: This should also be available as a `try_from_slice`

filmor (changes requested):

rustler/src/types/binary.rs
hauleth: I just applied the same logic as it is in `FromIterator` implementation. I think that both may be useful, as in most cases you cannot do anything with allocation error anyway. I will add `TryFrom` as well.

rustler/src/types/binary.rs
hauleth: I am not sure about having multiple ways to do the same thing with no extra value there. Just like we do not have `from_iterator` or similar. But I can add one if wanted.

rustler/src/types/binary.rs
filmor: Hmm. Rust itself is not super clear about this. `Vec<u8>` implements `From<&[u8]>` as well. And we don't have an additional error case to that.

hauleth: We can have both, that is not an issue. However my question is whether there is even a sense in trying to handle allocation errors, as these are pretty much impossible to handle.

rustler/src/types/binary.rs
hauleth: I have created #759 to discuss how to handle that, because there currently is no error that I could return from `TryFrom`.

rustler_tests/lib/rustler_test.ex
```diff
@@ -96,6 +96,12 @@ defmodule RustlerTest do
   def decode_iolist(_), do: err()
   def first_four_bytes_of_iolist(_), do: err()
 
+  def array_into_binary(), do: err()
```
filmor, line 99: Please add tests that actually call these.

rustler/src/types/binary.rs
```diff
@@ -262,6 +262,17 @@ impl FromIterator<u8> for OwnedBinary {
     }
 }
 
+impl<T: AsRef<[u8]>> From<T> for OwnedBinary {
+    fn from(data: T) -> Self {
```
filmor, line 274: I would still like this one to be available as `fn from_slice(data: impl AsRef<[u8]>) -> Self` on `OwnedBinary`. I find `let bin = OwnedBinary::from_slice(data)` more ergonomic than `let bin: OwnedBinary = data.into()`. The allocation error can stay as it is for now, until we have come to a conclusion on #761.

filmor (changes requested):

rustler/src/types/binary.rs
hauleth: You can call it as `OwnedBinary::from(data)` via `From` trait. Just like it is in Rust stdlib for `String` or `Vec`. I think that it offers what you want without any additional code.

rustler_tests/lib/rustler_test.ex
hauleth: These were mostly compilation tests, but I can add them.

rustler/src/types/binary.rs
```diff
@@ -135,6 +135,14 @@ impl OwnedBinary {
         })
     }
 
+    /// Copies 'data''s data into a new `OwnedBinary`.
+    pub fn from_slice(data: impl AsRef<[u8]>) -> Self {
```
hauleth, line 139: With that name it should accept `&[u8]` instead of `impl AsRef<[u8]>` IMHO.

PR #758, opened 2026-08-24
Change hash_internal to return u64

rkallos:
[Commit ef0d257501 in erlang/otp](https://github.com/erlang/otp/commit/ef0d257501) changed `enif_hash()` to use `erts_internal_salted_hash()`, which returns a unsigned 64-bit integer, as well as accepting a 64-bit salt. This commit was released as part of OTP 27, and there was no NIF version bump, as the signature for `enif_hash` didn't actually change.

This PR changes the signature of hash_internal to reflect the reality that `enif_hash` produces values in 0..2^64-1. The [erl_nif documentation](https://github.com/erlang/otp/blob/master/erts/doc/references/erl_nif.md) is out of date.

I found this was needed when trying to hash Erlang terms in a NIF, and the chosen hash implementation (hashbrown) used the most significant 7 bits for a particular optimization, which were truncated before these changes.

Thank you all for Rustler. It was a pleasure writing my first Rust NIF <3

### DIFF
--- a/rustler/src/term.rs
+++ b/rustler/src/term.rs
@@ -102,15 +102,9 @@ impl<'a> Term<'a> {
     /// Non-portable hash function that only guarantees the same hash for the same term within
     /// one Erlang VM instance.
     ///
-    /// It takes 32-bit salt values and generates hashes within 0..2^32-1.
-    pub fn hash_internal(&self, salt: u32) -> u32 {
-        unsafe {
-            enif_hash(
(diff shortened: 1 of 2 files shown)


filmor:
Thank you :)

Judging from a quick Github search, you may well be the only one who uses this function directly, so I don't think this should create issues for anyone. Will add it to the upgrade docs anyhow.

PR #456, opened 2022-05-26
Adds Env.whereis_pid

Qqwy:
I had the requirement to use `whereis_pid` in my code.
However, until now, `enif_whereis_pid` was exposed by `rustler_sys` but not wrapped by `rustler`; there was no way to build a new LocalPid from the returned ErlNifPid (without transmuting).

This PR adds a safe wrapper around it.

- [x] Implementation
- [x] Documentation
- [x] Tests

Please let me know what you think about the signature and the implementation.

PR #768, opened 2026-09-22
Drop rustler::wrapper

filmor:
Instead of using the intermediate half-safe function layer, we use the
`enif_*` functions now directly in the high level functions.


rubas:
exmpeg used `rustler::wrapper::NIF_ENV` only to keep the calling NIF's env in a struct without a lifetime parameter. It sends progress messages and checks caller liveness from a dirty NIF. We now store `Env<'a>` directly (rubas/exmpeg#86), so we do not need `wrapper`. Dropping it is fine for us.

Qqwy:
https://github.com/rusterlium/rustler/pull/456 made the usage of rustler::wrapper superfluous for `rustler_elixir_fun` but I never ended up upstreaming the change until it was merged. @filmor was quicker than me to create such a PR now after opening this issue (https://github.com/Qqwy/elixir-rustler_elixir_fun/pull/13). Thanks a lot for your help, and thank you very very much for your excellent work on Rustler!

akoutmos:
Thanks for the heads up and for all your hard work! Made the appropriate changes in Explorer https://github.com/elixir-explorer/explorer/pull/1164. When CI goes green I'll merge it in!

